PR #15, opened 2020-02-26
Simplified Verification for Referenced Parameters / Variables. (Fix for 5308)

StartAutomating:
Fixing by simplifying: no longer attempting to match a reference before the parameter/variable start

### DIFF
--- a/arm-ttk/testcases/deploymentTemplate/Parameters-Must-Be-Referenced.test.ps1
+++ b/arm-ttk/testcases/deploymentTemplate/Parameters-Must-Be-Referenced.test.ps1
@@ -9,26 +9,37 @@
 )
 
 <# REGEX
-- start with "
-- 0 or more whitespace
-- open bracket for expression [
-- any number of chars, the reference can be anywhere in the expression
+- 
 - parameters
-- 0 or more whitespace
-- open paren (
-- 0 or more whitespace
-- opening '
+- whitespace
+- ('
+- whitespace
+- <ParameterName>
+- whitespace
+- ')
 
-An expression could be: "[ concat ( parameters ( 'test' ), ...)]"
+An expression could be: "[ concat ( variables ( 'test' ), ...)]"
 
 #>
-
+$exprStrOrQuote = [Regex]::new('(?<!\\)(?>"\s{0,}\[|")', 'RightToLeft')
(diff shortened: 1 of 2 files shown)


arm-ttk/testcases/deploymentTemplate/Parameters-Must-Be-Referenced.test.ps1
```diff
@@ -27,7 +27,7 @@ foreach ($parameter in $TemplateObject.parameters.psobject.properties) {
 
     if (!($parameter.name.startswith('__'))) {
         
-        if ($TemplateText -notmatch "(?s)`"\s{0,}\[.*?parameters\s{0,}\(\s{0,}'$($Parameter.Name)'") {
+        if ($TemplateText -notmatch "parameters\s{0,}\(\s{0,}'$($Parameter.Name)'\s{0,}\)") {
```
bmoore-msft: need to make sure that we are looking for the "[ char combination otherwise it's not actually using the param (will be treated as a string).  Note that " and [ could be surrounded by any type of white space

bmoore-msft:
double check that the regex is correct too - pipe char looks out of place

bmoore-msft:
1) try fixing the original expression - look at regex options that span line breaks
2) could also remove line breaks from the text we test but this may cause a problem with reporting the line number
3) fix this PR to work (see comment above)

try in the order specified above...

arm-ttk/testcases/deploymentTemplate/Parameters-Must-Be-Referenced.test.ps1
```diff
@@ -9,26 +9,37 @@
...
 
 #>
-
+$exprStrOrQuote = [Regex]::new('(?<!\\)(?>"\s{0,}\[|")', 'RightToLeft')
 foreach ($parameter in $TemplateObject.parameters.psobject.properties) {
 
     if (!($parameter.name.startswith('__'))) {
         
-        if ($TemplateText -notmatch "(?s)`"\s{0,}\[.*?parameters\s{0,}\(\s{0,}'$($Parameter.Name)'") {
+        if ($TemplateText -notmatch ) {
```
bmoore-msft, line 29: this seems bad

PR #17, opened 2020-03-04
Template blanks unit tests

StartAutomating:


bmoore-msft:
closing looks like we may have had some branching issues since this doesn't contain the fix (just the unit test)

PR #18, opened 2020-03-04
Adding Check for RegEx Length Constraints (4064)

StartAutomating:


arm-ttk/testcases/CreateUIDefinition/Textboxes-Are-Well-Formed.test.ps1
```diff
@@ -9,6 +9,8 @@ $CreateUIDefinitionObject
 
 $allTextBoxes = $CreateUiDefinitionObject | Find-JsonContent -Key type -value Microsoft.Common.TextBox
 
+$LengthConstraint = [Regex]::new('(?<!\\)\{(?<Min>\d+),(?<Max>\d+)\}')
```
bmoore-msft: Translate for me - what is this checking?  Should be:

if there is a length constraint, the minimum length must be 12 chars

I don't see where we catch "12"

StartAutomating: This is checking that a constraint exists.  the "must be at least 12" logic was related to constraints on password boxes (this would be a separate pr, but would probably use the same Regex to find the length constraints). 

This test is making sure that the regex for any textbox implies a length.

bmoore-msft:
on regular textboxes we don't have a length requirement - only on password boxes

PR #34, opened 2020-04-08
Framework tests 4068

StartAutomating:
Pester unit testing for arm-ttk, and sample tests for DeploymentTemplate-Schema-Is-Correct


unit-tests/arm-ttk.tests.ps1
```diff
@@ -0,0 +1,201 @@
...
+    These tests make sure arm-ttk is working properly, and are not to be confused with the validation within arm-ttk.
+.Notes
+
+    The majority of tests are implemented in a parallel directory structure to the validation in arm-tttk.
+    
+    That is, for each test file in deploymentTemplate, a test directory should exist beneath this location.
+
+    For example, given the arm-ttk validation rule in:
+
+        deploymentTemplate\adminUsername-Should-Not-Be-A-Literal.test.ps1
```
bmoore-msft, line 16: can we make this the path from the root

/arm-ttk/testcases/deploymentTemplate...

unit-tests/arm-ttk.tests.ps1
```diff
@@ -0,0 +1,201 @@
...
+
+    There should be a test data directory:
+        
+        adminUsername-Should-Not-Be-A-Literal
```
bmoore-msft, line 20: Start from the root

/unit-tests/adminUsername-Should-Not-Be-A-Literal

unit-tests/arm-ttk.tests.ps1
```diff
@@ -0,0 +1,201 @@
...
+
+    
+    This will map to a describe block named deploymentTemplate\adminUsername-Should-Not-Be-A-Literal
+
+    This directory should contain two subfolders:
+
+    * Pass
```
bmoore-msft, line 27: /unit-tests/adminUsername-Should-Not-Be-A-Literal/Pass
/unit-tests/adminUsername-Should-Not-Be-A-Literal/Fail

unit-tests/arm-ttk.tests.ps1
```diff
@@ -0,0 +1,201 @@
...
+    ### The Pass Folder 
+    The Pass folder can contain one or more JSON files.  Running these rules on these files should produce no errors.
+    
+    The Pass folder may also contain one or more .path.txt files.  These will contain a relative path to a JSON file that should produce no errors.
+
+    ### The Fail Folder
+
+    The Fail folder may also contain one or more JSON files.  These JSON files are expected to produce errors.
+
+    Each JSON file may have a corresponding .should.be.ps1 or .should.be.txt
```
bmoore-msft, line 39: this is optional and we should explain why I would want to do it

unit-tests/adminUsername-Should-Not-Be-A-Literal/Fail/Literal-Admin-Username.json
```diff
@@ -0,0 +1,29 @@
+{
```
bmoore-msft, line 1: let's reduce this json to what's minimally needed and sort according to best practices

unit-tests/VM-Size-Should-Be-A-Parameter/Fail/No-Hardware-Profile.json
```diff
@@ -0,0 +1,26 @@
+{
```
bmoore-msft, line 1: probably don't need a case where hardware profile is missing, since it will fail to deploy... so just a failure case where it's not an expression for now

unit-tests/Common/Pass/100-marketplace-sample/metadata.json
```diff
@@ -0,0 +1,11 @@
+{
```
bmoore-msft, line 1: remove this file

unit-tests/Common/Pass/100-marketplace-sample/azuredeploy.parameters.json
```diff
@@ -0,0 +1,12 @@
+{
```
bmoore-msft, line 1: remove this file

Issue #51, opened 2020-04-30
Parameters-Must-Be-Referenced.test bug (and variables)


bmoore-msft:
Run the test on this sample:

https://github.com/Azure/azure-quickstart-templates/blob/master/301-nested-vms-in-virtual-network/azuredeploy.json

it flags hostAdminPassword as not being used but it is - VS Code knows it is, so doesn't seem to be anything in the file.

bmoore-msft:
variables test is flagging this one: 

https://github.com/Azure/azure-quickstart-templates/tree/master/101-vm-secure-password

Doesn't seem related to casing...

PR #52, opened 2020-04-30
Parameters must be referenced (https://github.com/Azure/arm-ttk/issues/51)

StartAutomating:
Fixing https://github.com/Azure/arm-ttk/issues/51 
Adding documentation / unit tests

### DIFF
--- a/arm-ttk/testcases/deploymentTemplate/Parameters-Must-Be-Referenced.test.ps1
+++ b/arm-ttk/testcases/deploymentTemplate/Parameters-Must-Be-Referenced.test.ps1
@@ -1,36 +1,49 @@
-﻿param(
+﻿<#
+.Synopsis
+    Ensures that parameters are referenced
+.Description
+    Ensures that all Azure Resource Manager Template
+#>
+param(
+    # The Template Object
     [Parameter(Mandatory = $true, Position = 0)]
     [PSObject]
     $TemplateObject,
 
+    # The Template JSON Text
     [Parameter(Mandatory = $true, Position = 0)]
     [PSObject]
     $TemplateText
 )
 
-<# REGEX
-- 
-- parameters
-- whitespace
-- ('
-- whitespace
-- <ParameterName>
-- whitespace
-- ')
-
-An expression could be: "[ concat ( variables ( 'test' ), ...)]"
-
-#>
 $exprStrOrQuote = [Regex]::new('(?<!\\)(?>"\s{0,}\[")', 'RightToLeft')
 foreach ($parameter in $TemplateObject.parameters.psobject.properties) {
-    $findVar = [Regex]::new("parameters\s{0,}\(\s{0,}'$($Parameter.Name)'\s{0,}\)")
-    $foundRefs = @($findVar.Matches($TemplateText))
-    if (-not $foundRefs) {
+    # If the parameter name starts with two underscores,
+    if ($parameter.Name -like '__*') { continue } # skip it.
+
+
+    # Create a Regex to find the var
+    $findVar = [Regex]::new(@"
+parameters           # the parameters keyword
+\s{0,}               # optional whitespace
+\(                   # opening parenthesis
+\s{0,}               # more optional whitespace
+(?:'|\\u0027)        # either a single quote or an escaped single quote
+$($Parameter.Name)   # the parameter name
+(?:'|\\u0027)        # either a single quote or an escaped single quote
+\s{0,}               # more optional whitespace
+\)                   # closing parenthesis
+"@,
+    # The Regex needs to be case-insensitive
+'Multiline,IgnoreCase,IgnorePatternWhitespace'
+)
+    $foundRefs = $findVar.Matches($TemplateText) # See if we found the variable
+    if (-not $foundRefs) { # If we didn't, error
         Write-Error -Message "Unreferenced parameter: $($Parameter.Name)" -ErrorId Parameters.Must.Be.Referenced -TargetObject $parameter
     } else {
-        foreach ($fr in $foundRefs) {
-            $foundQuote =$exprStrOrQuote.Match($TemplateText, $fr.Index)                
-            if ($foundQuote.Value -eq '"') {
+        foreach ($fr in $foundRefs) { # Walk thru each reference
+            $foundQuote =$exprStrOrQuote.Match($TemplateText, $fr.Index) # make sure we hit a [ before a quote
+            if ($foundQuote.Value -eq '"') { # if we don't, error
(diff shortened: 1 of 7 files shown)


arm-ttk/testcases/deploymentTemplate/Parameters-Must-Be-Referenced.test.ps1
```diff
@@ -1,36 +1,58 @@
...
+\(                   # opening parenthesis
+\s{0,}               # more optional whitespace
+(?:'|\\u0027)        # either a single quote or an escaped single quote
+$($Parameter.Name)   # the parameter name
+(?:'|\\u0027)        # either a single quote or an escaped single quote
+\s{0,}               # more optional whitespace
+\)                   # closing parenthesis
+"@,
+    # The Regex needs to be case-insensitive
+'Multiline,IgnoreCase,IgnorePatternWhitespace'
```
bmoore-msft, line 47: will this be the case with all regexes?  thinking if we just toLowered the source, that would be easier...

unit-tests/Parameters-Must-Be-Referenced/Pass/Different-Case-In-Reference.json
```diff
@@ -0,0 +1 @@
+﻿{"$schema":"https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#","parameters":{"myParameter":{"type":"string"}},"variables":{"foo":"[ parameters ( 'myParameter') ]"}}
```
bmoore-msft, line 1: format the json on all the tests (e.g. SHIFT+ALT+F in VS Code)

arm-ttk/testcases/deploymentTemplate/Parameters-Must-Be-Referenced.test.ps1
```diff
@@ -1,36 +1,58 @@
...
+    if ($parameter.Name -like '__*') { continue } # skip it.
+
+
+    # Create a Regex to find the var
+    $findVar = [Regex]::new(@"
+parameters           # the parameters keyword
+\s{0,}               # optional whitespace
+\(                   # opening parenthesis
+\s{0,}               # more optional whitespace
+(?:'|\\u0027)        # either a single quote or an escaped single quote
```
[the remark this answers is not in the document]
StartAutomating: In this case, no, because we're working off of an object that's converted back into JSON, and non-core PowerShell will annoyingly encode it as such.

We're allowing for both possibilities.

IMHO we should only have to do this in the cases where we ConvertTo-JSON within our test, and otherwise allow only single quotes.  Though I believe technically either would work (you tell me).

arm-ttk/testcases/deploymentTemplate/Parameters-Must-Be-Referenced.test.ps1
StartAutomating: It won't be the case when we're using -match (that applies it normally). It will be the case whenever we construct it ourselves.  If we move all the Regex into a common location we can enforce this on module load.

The IgnorePatternWhitespace is required for readability.  Multiline actually is _mostly_ noise here (technically, it should allow the whitespace to span a line break).

unit-tests/Parameters-Must-Be-Referenced/Pass/Different-Case-In-Reference.json
StartAutomating: done

arm-ttk/testcases/deploymentTemplate/Parameters-Must-Be-Referenced.test.ps1
StartAutomating: As for to ToLower suggestion, I'm conflicted.  Everything I know of JSON indicates we shouldn't muck with casing, and a uniform ToLower on TemplateText would create confusing looking responses (e.g. "Parameter 'adminusername' is not referenced" where the JSON has "adminUsername").

On the other hand, everything I know about ARM indicates we should be able to do this without serious consequences, at least from a pure validation standpoint.

I'm more inclined to leave it as-is and ToLower() from time to time if needed, or to start pre-loading our Regexes and defining them in module subdirectories, under the general principle of "Better Explicit than Terse"

arm-ttk/testcases/deploymentTemplate/Parameters-Must-Be-Referenced.test.ps1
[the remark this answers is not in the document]
bmoore-msft: I'd like to get rid of the encoding, it's not readable and won't be obvious as to when we need it and when we don't - I thinking I'm ok relying on PS Core at this point.

arm-ttk/testcases/deploymentTemplate/Parameters-Must-Be-Referenced.test.ps1
bmoore-msft: Ack on the error message, though we might be able to solve that another way...

Nothing that is ARM specific, is case sensitive - e.g. property names, expression syntax... and if we retain case then all tests and all testcase have to account for it, so seems like that's the more difficult path

arm-ttk/testcases/deploymentTemplate/Parameters-Must-Be-Referenced.test.ps1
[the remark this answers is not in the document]
StartAutomating: I've changed the Regex to use single quotes, and updated the ConvertTo-Json line to do the un-escaping.  This will continue to work in both cases.

arm-ttk/testcases/deploymentTemplate/Parameters-Must-Be-Referenced.test.ps1
```diff
@@ -16,37 +16,39 @@ param(
     $TemplateText
 )
 
-$TemplateObjectCopy = [PSObject]::new()
-foreach ($prop in $TemplateObject.psobject.properties) {
-    if ($prop.Name -eq 'Parameters') { continue }
+# We need the -TemplateText, but without parameters, so 
+$TemplateObjectCopy = [PSObject]::new() # create a shallow copy
+foreach ($prop in $TemplateObject.psobject.properties) { # of every property 
+    if ($prop.Name -eq 'Parameters') { continue } # except 'Parameters'.
```
bmoore-msft, line 22: let's change our approach here and rather than make a copy of the stuff without params (or vars) let's just use a refcount of > 1... iow, we should find the text in the template more than once.

PR #56, opened 2020-05-06
Copy variable support (4411)

StartAutomating:


arm-ttk/testcases/deploymentTemplate/Variables-Must-Be-Referenced.test.ps1
```diff
@@ -14,31 +14,29 @@ param(
     $TemplateText
```
bmoore-msft, line 14: as we're doing with the parameter test - let's use the approach of looking for the text match > 1 time... it doesn't look like we make a copy here like we did in the other test (but probably should have) but let's do them the same way regardless

arm-ttk/testcases/deploymentTemplate/Variables-Must-Be-Referenced.test.ps1
```diff
@@ -14,31 +14,29 @@ param(
...
-# TODO: Need to properly check for variable copy, see: https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-create-multiple#variable-iteration
-
-$exprStrOrQuote = [Regex]::new('(?<!\\)(?>"\s{0,}\[|")', 'RightToLeft')
-
+$exprStrOrQuote = [Regex]::new('(?<!\\)[\[\"]', 'RightToLeft')
 foreach ($variable in $TemplateObject.variables.psobject.properties) {
     
     # TODO: if the variable name is "copy": we need to loop through the array and pull each var and check individually
     
     if ($variable.name -ne 'copy' ) {
```
bmoore-msft, line 22: there are multiple ways copy can manifest in variables, see: https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/copy-variables

copy can be a direct descendent of the variables object or a grandchild - if you want to enter an issue for this and fix later that should be fine

PR #78, opened 2020-05-27
updates to check for nested expressions + unit tests

bmoore-msft:


bmoore-msft:
need to adopt regex pattern from other tests

PR #89, opened 2020-06-11
Deployment resources must not be debug (4144)

StartAutomating:


StartAutomating:
Added /Pass/DebugSettingsNone.json

PR #91, opened 2020-06-14
DependsOn must not start with concat (4163)

StartAutomating:


unit-tests/DependsOn-Must-Not-Be-Conditional/Fail/DependsOn-Starts-With-Concat.json
```diff
@@ -0,0 +1,8 @@
+﻿{
```
bmoore-msft, line 1: let's add a case that should pass

dependsOn: "[resourceId('Microsoft.Storage/storageAccounts', concat('foo', copyIndex()))]"

PR #97, opened 2020-06-18
Hardcoded uri checks (Azure DevOps Issues 4759, 4067, 4072)

StartAutomating:


unit-tests/DeploymentTemplate-Must-Not-Contain-Hardcoded-Uri/Fail/Hardcoded-Uri.json
```diff
@@ -0,0 +1,6 @@
+{
```
bmoore-msft, line 1: add some cases for the other/full scenarios

http://foo.com
https://foo.com
.foo.bar.com
foo.bar.com
also make sure that some are in expressions, e.g.
concat('https://', 'blob.core.windows.net')

Passing tests:
e.g. "AdTenant": "Common"

StartAutomating: I added some additional tests, and support for any Suffix properties that can be a URI.

I noticed that blob.core.windows.net isn't in the list we can infer.

I think we may want to add a parameter containing a customizable blacklist, so that this particular URI (blob.core.windows.net) is flagged.

arm-ttk/testcases/deploymentTemplate/DeploymentTemplate-Must-Not-Contain-Hardcoded-Uri.test.ps1
```diff
@@ -0,0 +1,46 @@
...
+    Ensures Deployment Templates do not use hardcoded URLs
+.Description
+    Ensures Deployment Templates do not use Hardcoded URLs found in the AllEnvironments cache.
+#>
+param(
+# The template text.
+[string]
+$TemplateText,
+
+# The cached date of all environments (contained in /cache/AllEnvironments.cache.json)
```
bmoore-msft, line 12: date or data?

unit-tests/DeploymentTemplate-Must-Not-Contain-Hardcoded-Uri/Fail/Hardcoded-Uri.json
bmoore-msft: double check - core.windows.net is in the list so this should have flagged

arm-ttk/testcases/deploymentTemplate/DeploymentTemplate-Must-Not-Contain-Hardcoded-Uri.test.ps1
```diff
@@ -0,0 +1,46 @@
...
+[PSObject[]]
+$AllEnvironments
+)
+
+$uniqueHosts = # Walk thru all environments
```
bmoore-msft, line 17: Let's change this to copy/paste the blacklist into the test itself, like we do everywhere else... then we can do a literal string match for all files (not just templates).

It's not ideal, but when we solve this we should do it for all tests.  This will simplify the test code quite a bit as the tradeoff.

Then we can just look for a match in TemplateText (excluding comments)...

arm-ttk/testcases/deploymentTemplate/DeploymentTemplate-Must-Not-Contain-Hardcoded-Uri.test.ps1
```diff
@@ -0,0 +1,46 @@
...
+    @(foreach ($envData in $AllEnvironments) {
```
bmoore-msft, line 18: Here are the things we want from the public cloud, also need it for the other clouds.

"management.core.windows.net",
"gallery.azure.com",
"management.core.windows.net",
"management.azure.com",
"database.windows.net",
"core.windows.net",
"login.microsoftonline.com",
"graph.windows.net",
"graph.windows.net",
"trafficmanager.net",
"vault.azure.net",
"datalake.azure.net",
"azuredatalakestore.net",
"azuredatalakeanalytics.net",
"vault.azure.net",
"api.loganalytics.io",
"api.loganalytics.iov1",
"asazure.windows.net",
"region.asazure.windows.net",
"api.loganalytics.iov1",
"api.loganalytics.io",
"asazure.windows.net",
"region.asazure.windows.net",
"batch.core.windows.net"

PR #100, opened 2020-06-18
Reference api version check (4162)

StartAutomating:


bmoore-msft:
check the workitem for more details of the test

PR #103, opened 2020-06-18
CreateUIDefinition blanks test (4596)

StartAutomating:


unit-tests/CreateUIDefinition-Must-Not-Have-Blanks/Fail/CreateUIDefinition-With-EmptyArray/createUiDefinition.json
```diff
@@ -0,0 +1,8 @@
+{
```
bmoore-msft, line 1: this should be a pass test (we added the exception)

PR #112, opened 2020-06-22
Password box min length 4084

StartAutomating:


unit-tests/PasswordBoxes-Must-Have-Min-Length/Fail/No-Regex-Constraint.json
```diff
@@ -0,0 +1,26 @@
+﻿{
```
bmoore-msft, line 1: this test should pass - when no regex is provided the default is used and the default has the proper minlength

StartAutomating: Fixed, and updated the test to no longer fail under that condition.

Issue #122, opened 2020-06-30
Virtual-Machines-Should-Not-Be-Preview test throws odd false positive


bmoore-msft:
Run the tests on this sample:

https://github.com/Azure/azure-quickstart-templates/tree/master/101-machine-learning-compute-attach-hdi

That's about all I can figure out about this one...

PR #165, opened 2020-07-16
Limit check to virtualMachines and virtualMachineScaleSets

hallihan:
Currently the check will trigger a false positive if a variable or parameter is named 'storageProfile'.  Add a check to filter the results from Find-JsonContent to only include nodes where the parent object is a VM or VMSS.  Testing done: Tested script locally with vm and vmss templates manually modified to include "-preview" in the version field.  This might introduce a limitation of not finding cases where a variable named 'storageProfile' is defined and actually contains an 'imageReference', which is later used as the value for the 'storageProfile' field of a VM or VMSS.  This should also address https://github.com/Azure/arm-ttk/issues/122 which I'm guessing is happening because the prereq template for the HDI cluster includes a storageProfile property (https://github.com/Azure/azure-quickstart-templates/blob/master/101-machine-learning-compute-attach-hdi/prereqs/prereq.azuredeploy.json#L197) as will any other template that deploys HDInsight.

### DIFF
--- a/arm-ttk/testcases/deploymentTemplate/Virtual-Machines-Should-Not-Be-Preview.test.ps1
+++ b/arm-ttk/testcases/deploymentTemplate/Virtual-Machines-Should-Not-Be-Preview.test.ps1
@@ -10,7 +10,7 @@ param(
 $TemplateObject
 )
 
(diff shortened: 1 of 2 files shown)


bmoore-msft:
#122 - thanks Rick... also added a test case for this ;)

PR #188, opened 2020-09-03
Test refactor

StartAutomating:
Refactoring unit tests to be able to run all at once or per-directory.


unit-tests/arm-ttk.tests.ps1
```diff
@@ -162,7 +162,7 @@ $(if ($ttkError.FullyQualifiedErrorID -notlike 'Microsoft.PowerShell*') {
     }
 }
 
-
+<#
```
bmoore-msft, line 165: 2 things:

1) let's keep this file/scenario working

2) let's add a MD file that documents the intended uses for our repo, e.g. how to run a single suite of tests and how to run all tests

RunPester.yml
```diff
@@ -0,0 +1,89 @@
...
+          [string]
+          $ModulePath,
+          # The Pester max version.  By default, this is pinned to 4.99.99.
+          [string]
+          $PesterMaxVersion = '4.99.99'
+          )
+          
+          $orgName, $moduleName = $env:GITHUB_REPOSITORY -split "/"
+          if (-not $ModulePath) {
+              $orgName, $moduleName = $env:GITHUB_REPOSITORY -split "/"
```
bmoore-msft, line 68: can we remove this since it's 2 lines above (unconditionally?)

bmoore-msft (changes requested):

PR #206, opened 2020-10-02
Add support for jsonc files - fixes #200

### LINKED ISSUE #200: No support for jsonc files
On a quick scan of this repo there's no support at all for using jsonc files, which whilst uncommon compared to normal json files should be allowed as ARM is happy with accepting deployments that use jsonc files.

kilasuit:


bmoore-msft (changes requested):
Changes look good - can we add some test cases to make sure this is working as expected...  I think you can copy any of the existing ones and put them side by side.  Make sure we have a pass and fail.

bmoore-msft:
Closing abandoned request - please reopen if you can address the feedback.

Issue #193, opened 2020-09-30
false positive on parameter reference


bmoore-msft:
See this template for repro: https://github.com/Azure/azure-quickstart-templates/tree/master/101-logic-app-create

The testuri parameter is flagged as not being used but it is used in the workflow definition

Splaxi:
I found the same today. Do you want a hand to fix it?

bmoore-msft:
Please ;)

Splaxi:
The real issue is in fact that the ARM template parameter is named the same as the WorkFlow parameter (Logic App).

So because the language used for ARM templates and the language used for Logic App resembles on so many levels, including parameters, the regex is getting false positives.

From a deployment perspective, the workflow is a bit of wild horse. It is fully supported that we use a ARM parameter inside the workflow definition, while also defining workflow parameters, that we use later in the workflow definition.

When referencing workflow parameter, they must always by enclosed in double quotes, and a @ character, and sometimes inside a set of brackets {}, depending on how advanced the expression is.

Maybe a solution to be to write a regex that will ignore the following:
"@parameters('ParmName')"
"@funtion1(parameters('ParmName'))"
"@{parameters('ParmName')}"
"@{function1(parameters('ParmName'))}"

funtion1 is just to show that there are several builtin functions in the workflow definition that can consume the parameter value.

"@float(string(parameters('ParmName')))" is a valid statement.
"adfasdf@{float(string(parameters('ParmName')))}adsfaf" is a valid statement.

Splaxi:
Looking a bit more into the regex, as there are 2 regex in play here:

$exprStrOrQuote = [Regex]::new('(?<!\\)[\[\"]', 'RightToLeft')

This is the just saying: Find **[** or **"**, from right to left. If I change it to: $exprStrOrQuote = [Regex]::new('(?<!\\)[\[\"@]', 'RightToLeft'), it seems in my testing that it solves the issue.

I'll do a PR right away and let you take it for a spin.

bmoore-msft:
Ok, interesting revelations from #212...  TLDR;

- The syntax of ```@parameters()``` or ```@something(parameters()``` should not be counted as a reference of an ARM template parameter.
- the current test is throwing on any invalid reference() even if there are valid references - that's the current design but a discussion of whether it's correct is warranted
- if the current design is correct (and we should throw on any invalid reference even when there are valid) then we have to figure how to except certain cases if and only if there is also valid reference from ARM.

[...]

PR #212, opened 2020-10-13
Fix workflow issue

Splaxi:
Trying to solve the #193 

It boils down to the workflow definition can use parameter in the same style as ARM templates, and if you use the same name for the workflow parameters as the ARM parameter, the current test fails.

### DIFF
--- a/arm-ttk/testcases/deploymentTemplate/Parameters-Must-Be-Referenced.test.ps1
+++ b/arm-ttk/testcases/deploymentTemplate/Parameters-Must-Be-Referenced.test.ps1
@@ -16,7 +16,7 @@ param(
     $TemplateText
 )
 
-$exprStrOrQuote = [Regex]::new('(?<!\\)[\[\"]', 'RightToLeft')
+$exprStrOrQuote = [Regex]::new('(?<!\\)[\[\"@]', 'RightToLeft')
 foreach ($parameter in $TemplateObject.parameters.psobject.properties) {
     # If the parameter name starts with two underscores,
     if ($parameter.Name -like '__*') { continue } # skip it.
--- a/arm-ttk/testcases/deploymentTemplate/Parameters-Must-Be-Referenced.test.ps1
+++ b/arm-ttk/testcases/deploymentTemplate/Parameters-Must-Be-Referenced.test.ps1
@@ -42,7 +42,7 @@ $escapedName  # the parameter name
         Write-Error -Message "Unreferenced parameter: $($Parameter.Name)" -ErrorId Parameters.Must.Be.Referenced -TargetObject $parameter
     } else {
         foreach ($fr in $foundRefs) { # Walk thru each reference
-            $foundQuote =$exprStrOrQuote.Match($TemplateText, $fr.Index + 1) # make sure we hit a [ before a quote
+            $foundQuote = $exprStrOrQuote.Match($TemplateText, $fr.Index + 1) # make sure we hit a [ before a quote
             if ($foundQuote.Value -eq '"') { # if we don't, error
                 Write-Error -Message "Parameter reference is not contained within an expression: $($Parameter.Name)" -ErrorId Parameters.Must.Be.Referenced.In.Expression -TargetObject $parameter
             }
--- a/unit-tests/Parameters-Must-Be-Referenced/Pass/Workflow-Using-Same-Named-Parameters.json
+++ b/unit-tests/Parameters-Must-Be-Referenced/Pass/Workflow-Using-Same-Named-Parameters.json
@@ -1,0 +1,68 @@
+{
+    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "logicAppName": {
+            "type": "string",
+            "metadata": {
+                "description": "The name of the logic app to create."
+            }
+        },
+        "testUri": {
+            "type": "string",
+            "defaultValue": "https://status.azure.com/en-us/status/",
+            "metadata": {
+                "description": "A test URI"
+            }
+        },
+        "location": {
+            "type": "string",
+            "defaultValue": "[resourceGroup().location]",
+            "metadata": {
+                "description": "Location for all resources."
+            }
+        }
+    },
+    "variables": {},
+    "resources": [
+        {
+            "type": "Microsoft.Logic/workflows",
+            "apiVersion": "2019-05-01",
+            "name": "[parameters('logicAppName')]",
+            "location": "[parameters('location')]",
+            "tags": {
+                "displayName": "LogicApp"
+            },
+            "properties": {
+                "definition": {
+                    "$schema": "https://schema.management.azure.com/providers/Microsoft.Logic/schemas/2016-06-01/workflowdefinition.json#",
+                    "contentVersion": "1.0.0.0",
+                    "parameters": {
+                        "testUri": {
+                            "type": "string",
+                            "defaultValue": "[parameters('testUri')]"
+                        }
+                    },
+                    "triggers": {
+                        "recurrence": {
+                            "type": "recurrence",
+                            "recurrence": {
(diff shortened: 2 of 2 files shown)


Splaxi:
Hi @bmoore-msft 

Looking at the output from the checks / tests, it seems that you have some pester issues with the unit tests that should fail for the ResourceIds-should-not-contain

bmoore-msft:
thx - I added a PR for the resourceId test fixes...

re: this test - it's not clear to me what the test is checking?  If I have this syntax in my template:

```@parameter('testUri')```

My understanding is that is referencing the workflow parameter and not the template parameter.

And then of course the template syntax:

```[ ... parameters('testUri') ... ]```

Would count as a reference to the template parameter.

If I have the right, I think we need more unit tests around it... and not sure I understand the edition of the @ in the regex - but let's make sure I understand this right first ;)

**Comment by Splaxi** [comment]
I'll see if I can make it clear.

The ARM template for the unit test: https://github.com/Splaxi/arm-ttk/blob/fix-workflow-issue/unit-tests/Parameters-Must-Be-Referenced/Pass/Workflow-Using-Same-Named-Parameters.json

It has an ARM template parameter called **testUri**:
https://github.com/Splaxi/arm-ttk/blob/6276251452a454a2e4e9e9f70a7812cd9b944ae6/unit-tests/Parameters-Must-Be-Referenced/Pass/Workflow-Using-Same-Named-Parameters.json#L11-L17

image

The workflow definition also has a parameter section, where is has a parameter the is named 100% the same:
https://github.com/Splaxi/arm-ttk/blob/6276251452a454a2e4e9e9f70a7812cd9b944ae6/unit-tests/Parameters-Must-Be-Referenced/Pass/Workflow-Using-Same-Named-Parameters.json#L40-L45

image

When the current Test-AzTemplate runs the **Parameters-Property-Must-Exist.test.ps1** test case, the following is happening:

1. It parses all ARM parameters, and loops/iterates with every single one of them with the regex:
https://github.com/Splaxi/arm-ttk/blob/6276251452a454a2e4e9e9f70a7812cd9b944ae6/arm-ttk/testcases/deploymentTemplate/Parameters-Must-Be-Referenced.test.ps1#L26-L39

image

What this will yield, is 3 matches:

- 1 match for the **logicAppName** parameter
```
Groups   : {0}
Success  : True
Name     : 0
Captures : {0}
Index    : 736
Length   : 26
Value    : parameters('logicAppName')
``` 
- 2 matches for the **testUri** parameter
```
Groups   : {0}
Success  : True
Name     : 0
Captures : {0}
Index    : 1270
Length   : 21
Value    : parameters('testUri')
```
and 
```
Groups   : {0}
Success  : True
Name     : 0
Captures : {0}
Index    : 1947
Length   : 21
Value    : parameters('testUri')
```

The match for index 1270 is:
https://github.com/Splaxi/arm-ttk/blob/6276251452a454a2e4e9e9f70a7812cd9b944ae6/unit-tests/Parameters-Must-Be-Referenced/Pass/Workflow-Using-Same-Named-Parameters.json#L40-L45

image

The match for index 1947 is:
https://github.com/Splaxi/arm-ttk/blob/6276251452a454a2e4e9e9f70a7812cd9b944ae6/unit-tests/Parameters-Must-Be-Referenced/Pass/Workflow-Using-Same-Named-Parameters.json#L58-L61

image

For each match found, we will run the next `$exprStrOrQuote` regex, which will look from right to left, for either **[** or **"**
https://github.com/Splaxi/arm-ttk/blob/6276251452a454a2e4e9e9f70a7812cd9b944ae6/arm-ttk/testcases/deploymentTemplate/Parameters-Must-Be-Referenced.test.ps1#L19

image

So for the match at index 1270 that is not a problem, because the first character we encounter is [, which is what we want. The result from the `$exprStrOrQuote` is: 
(`"defaultValue": "[parameters('testUri')]"`)
```
Groups   : {0}
Success  : True
Name     : 0
Captures : {0}
Index    : 1269
Length   : 1
Value    : [
```

For the match at index 1947, the story is different. The reason for that is that **inside** the workflow definition (Logic App - "language"/syntax ), we don't use the [] characters around parameters when we reference them (the internal ones, the ones that are declared inside the resource). So the result is:
(`"uri": "@parameters('testUri')"`)
```
Groups   : {0}
Success  : True
Name     : 0
Captures : {0}
Index    : 1945
Length   : 1
Value    : "
```
That yields the failed unit tests, because we should have a [ character and not a " character.

This is were the @ character comes into play. If our regex is looking for 3 characters [ " @, the last match will look like this:
(`"uri": "@parameters('testUri')"`)
```
Groups   : {0}
Success  : True
Name     : 0
Captures : {0}
Index    : 1946
Length   : 1
Value    : @
```

So stepping back a bit. Either we need to rewrite the big search regex that initially finds the match on index 1947, because it shouldn't be found in the first place. How to do that would be something with negative-lookback magic. Or we use this suggested solution (some would argue hack), were we recognize that we have a special case with workflow, that requires a special lookup regex for the validation.

I hope it makes things a bit more clear 🤞

**Comment by bmoore-msft** [comment]
It does help... To level set, the ```@parameters()``` syntax is only used for the logic workflow, not for the ARM template.  So that sytnax should be ignored or iow, not treated as a reference to the param.  So we don't actually what this change: https://github.com/Azure/arm-ttk/pull/212/files#diff-b6bd0934fae1999159cd89b86d3c57d9fd0b467be055e72d5bfd7352ed18294bR19

The problem seems to be when I find a not reference, we throw, even though we may also have a reference.  So the problem is more nefarious than I thought.  For example, this throws an error and it shouldn't.

```
{
  "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
  "contentVersion": "1.0.0.0",
  "parameters": {
    "testUri": {
      "type": "string"
    }
  },
  "variables": {
    "ref": "[parameters('testUri')]",
    "notRef": "parameters('testUri')"
  },
  "resources": []
}
```

This is also a bug in the variable test (I'm guessing) but that pattern is not common - you'd have to have a valid reference and a string match for invalid before it triggers, which is not common in the wild and usually indicative of something you do want to pay attention to... except in the case of a logic app.

Not immediately obvious to me how to fix, but I think we understand the problem better (or I do ;))

Splaxi:
Let me ask it like this.

Should one reference for any given parameter be enough to not fail or shall/must all references for a given parameter be validated? The latter is what your current regex is enforcing.

Just remember that `@parameter()` can be anything like `@float(parameter(''))` and `"dsafasdfasf@{string(float(parameter('')))}dfasdfaf"`

It seems that the current regex is greedy on purpose, to make sure we catch all "maybe" references for a given parameter, to ensure that it is inside an expression. If we move away from that, by changing the regex (somehow), we would fix the issue - but might hit more edge cases were we don't catch all references for the parameter that we would want to.

So I would argue that this change is an improvement, worthwhile to implement, because it solves the edge case of workflow / logicapps, and lets us keep the remaining greedy logic to find all other references of parameters from the ARM template. And maybe someday we find an even better solution, but for now we can support ARM templates, with workflow / logicapps, where the inner parameter is named exactly the same.

But I fully understand if you want to close this PR - it is your call anyways. I'm just trying to help.

bmoore-msft:
Ok, I think I see the disconnect.  The syntax of ```@parameters()``` or ```@something(parameters()``` should not be counted as a reference of an ARM template parameter.  It *is* a reference of a parameter in a logicApp workflow, but that's not what we're testing.  So adding @ to the regex is not the correct fix.

The problem described in #193 is that we throw on any invalid reference regardless of whether there is also a valid reference.  As long as there is at least one valid reference, the test should pass.  Also, it's not the regex that's causing that it's the loop in the references.

Now there is a discussion around whether we should flag any invalid reference, which has merit because it's an odd thing to do - and that's what caused the problem here, just not the way the test was designed.  If we were to do that, we'd have to special case scenarios like this...  because I would need to "discard" references that were not from the ARM part of the template, but are not invalid. 

So I think we should close this and open a new issue now that there's a better understanding of #193

PR #217, opened 2020-10-20
Adding concat as a valid expression

SebastianClaesson:
Concatenation of strings can be used when refering custom policies.

An example could be a custom policy created with a friendly name such as "Test-Policy1"
Then a concatenation of the policy could be SCOPE/providers/Microsoft.Authorization/policyDefinitions/Test-Policy1
Where the scope could be something like:
/providers/Microsoft.Management/managementGroups/ITManagementGroup

We use this in our ARM template and currently validation fails at this step, so either it should be in the exclude section or also allow concat as a valid expression.

### DIFF
--- a/arm-ttk/testcases/deploymentTemplate/IDs-Should-Be-Derived-From-ResourceIDs.test.ps1
+++ b/arm-ttk/testcases/deploymentTemplate/IDs-Should-Be-Derived-From-ResourceIDs.test.ps1
@@ -82,7 +82,8 @@ foreach ($id in $ids) { # Then loop over each object with an ID
         "clientId",
         "appId",
         "tenantId",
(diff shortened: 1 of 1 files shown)


bobbyangers:
you seem to have put it in the list of exceptions 'id' instead of the allowed functions

bmoore-msft:
concat() is the function that we're actually trying to flag....  for your two scenarios the property syntax would be:

```/providers/Microsoft.Management/managementGroups/ITManagementGroup```
should be
```tenantResourceId('Microsoft.Management/managementGroups', 'ITManagementGroup')```

Each resourceId format has it's own function that's appropriate for use...

So this would be a case the we do *not* want to support.

PR #218, opened 2020-10-22
Dev/workflow id skip

bobbyangers:
bypass id check within Microsoft.Logic/workflows

### DIFF
--- a/arm-ttk/testcases/deploymentTemplate/IDs-Should-Be-Derived-From-ResourceIDs.test.ps1
+++ b/arm-ttk/testcases/deploymentTemplate/IDs-Should-Be-Derived-From-ResourceIDs.test.ps1
@@ -44,11 +44,6 @@ foreach ($id in $ids) { # Then loop over each object with an ID
     if ($id.JsonPath -match '^(parameters|outputs)') {
         continue
     }
-	
-    #ID inside a Microsoft.Logic/workflows should be skipped
-    if ($id.JsonPath -match '\.properties(\[\d{1,2}\])?\.definition(\[\d{1,2}\])?\.') {
-        continue
-    }
 
     if ($myId -isnot [string] -and ($myId -as [float] -eq $null)) {
         if (-not $myId.Value) {
@@ -67,7 +62,34 @@ foreach ($id in $ids) { # Then loop over each object with an ID
         continue
     }
     $expandedId = Expand-AzTemplate -Expression $myId -InputObject $TemplateObject -Exclude Parameters # then expand it.
+
+    if ($expandedId -isnot [string]) { #if it happens to be an object property, skip it
+        continue
+    };
+
+     # Check that it uses one of the allowed expressions - can remove variables once Expand-Template does full eval of nested vars
+    # REGEX
+    # - 0 or more whitespace
+    # - @ 
+    # - expression must be 
+    #    - option 1: 
+    #       - opening {
+    #       - any expression in betwwen
+    #       - closing }
+    #    - option 2
+    #       - alpha function name      
+    #       - opening paren (
+    #       - closing paren )
+    #  Examples
+    #    Option 1  =>   @{coalesce(triggerOutputs().headers?['id'], guid())}
+    #    Option 2  =>   @body('Get_email_(V2)_-_Processing')?['id']
+
+    $exprMatch0 = "^\s{0,}@((\{.*\})|([a-bA-B]*\(.*\)))"
+    if ($expandedId -match $exprMatch0  ){
+        continue;
+    }
     
+
     # these are allowed for resourceIds
     $allowedExpressions = @(
         "extensionResourceId",
@@ -98,9 +120,7 @@ foreach ($id in $ids) { # Then loop over each object with an ID
     # - single quote on parameters and variables (resourceId first parameters may not be a literal string)
     #
     $exprMatch = "\s{0,}\[\s{0,}($($allowedExpressions -join '|' ))\s{0,}\(\s{0,}"
-
-    if ($expandedId -is [string] -and ` #if it happens to be an object property, skip it
-        $expandedId -notmatch $exprMatch  ){
+    if ($expandedId -notmatch $exprMatch  ){
             Write-Error "Property: `"$($id.propertyName)`" must use one of the following expressions for an resourceId property:
             $($allowedExpressions -join ',')
             Path: [$($id.JsonPath)]"`
--- a/arm-ttk/testcases/deploymentTemplate/IDs-Should-Be-Derived-From-ResourceIDs.test.ps1
+++ b/arm-ttk/testcases/deploymentTemplate/IDs-Should-Be-Derived-From-ResourceIDs.test.ps1
@@ -84,7 +84,7 @@ foreach ($id in $ids) { # Then loop over each object with an ID
     #    Option 1  =>   @{coalesce(triggerOutputs().headers?['id'], guid())}
     #    Option 2  =>   @body('Get_email_(V2)_-_Processing')?['id']
 
-    $exprMatch0 = "^\s{0,}@((\{.*\})|([a-bA-B]*\(.*\)))"
+    $exprMatch0 = "^\s{0,}@((\{.*\})|([\w0-9]+\(.+\)))"
     if ($expandedId -match $exprMatch0  ){
         continue;
     }
--- a/arm-ttk/testcases/deploymentTemplate/IDs-Should-Be-Derived-From-ResourceIDs.test.ps1
+++ b/arm-ttk/testcases/deploymentTemplate/IDs-Should-Be-Derived-From-ResourceIDs.test.ps1
@@ -44,6 +44,11 @@ foreach ($id in $ids) { # Then loop over each object with an ID
     if ($id.JsonPath -match '^(parameters|outputs)') {
         continue
     }
+	
+    #ID inside a Microsoft.Logic/workflows should be skipped
+    if ($id.JsonPath -match '\.properties\.definition') {
+        continue
+    }
 
     if ($myId -isnot [string] -and ($myId -as [float] -eq $null)) {
         if (-not $myId.Value) {
@@ -55,9 +60,6 @@ foreach ($id in $ids) { # Then loop over each object with an ID
             }
         }
     }
-
-    
-    
 
     # $myId = "$($id.id)".Trim() # Grab the actual ID,
     if (-not $myId) {
--- a/arm-ttk/testcases/deploymentTemplate/IDs-Should-Be-Derived-From-ResourceIDs.test.ps1
+++ b/arm-ttk/testcases/deploymentTemplate/IDs-Should-Be-Derived-From-ResourceIDs.test.ps1
@@ -104,6 +104,7 @@ foreach ($id in $ids) { # Then loop over each object with an ID
             Write-Error "Property: `"$($id.propertyName)`" must use one of the following expressions for an resourceId property:
             $($allowedExpressions -join ',')" `
              -TargetObject $id -ErrorId ResourceId.Should.Contain.Proper.Expression
+             Write-Error "Path: `"$($id.JsonPath)`"
     }
 }
 
--- a/arm-ttk/testcases/deploymentTemplate/IDs-Should-Be-Derived-From-ResourceIDs.test.ps1
+++ b/arm-ttk/testcases/deploymentTemplate/IDs-Should-Be-Derived-From-ResourceIDs.test.ps1
@@ -46,7 +46,7 @@ foreach ($id in $ids) { # Then loop over each object with an ID
     }
 	
     #ID inside a Microsoft.Logic/workflows should be skipped
-    if ($id.JsonPath -match '\.properties\.definition') {
+    if ($id.JsonPath -match '\.properties(\[\d{1,2}\])?\.definition(\[\d{1,2}\])?\.') {
         continue
     }
 
@@ -102,9 +102,9 @@ foreach ($id in $ids) { # Then loop over each object with an ID
     if ($expandedId -is [string] -and ` #if it happens to be an object property, skip it
         $expandedId -notmatch $exprMatch  ){
             Write-Error "Property: `"$($id.propertyName)`" must use one of the following expressions for an resourceId property:
-            $($allowedExpressions -join ',')" `
-             -TargetObject $id -ErrorId ResourceId.Should.Contain.Proper.Expression
-             Write-Error "Path: [$($id.JsonPath)]"
+            $($allowedExpressions -join ',')
+            Path: [$($id.JsonPath)]"`
+            -TargetObject $id -ErrorId ResourceId.Should.Contain.Proper.Expression
     }
 }
 
--- a/arm-ttk/testcases/deploymentTemplate/IDs-Should-Be-Derived-From-ResourceIDs.test.ps1
+++ b/arm-ttk/testcases/deploymentTemplate/IDs-Should-Be-Derived-From-ResourceIDs.test.ps1
@@ -30,12 +30,12 @@ foreach ($id in $ids) { # Then loop over each object with an ID
     # these properties are exempt, since they are not actually resourceIds
     $exceptions = @(
         "tenantId",
-        "workerSizeId", # Microsoft.Web/serverFarms
-        "serverFarmId", # Microsoft.Web/sites
+        "workerSizeId",     # Microsoft.Web/serverFarms
+        "serverFarmId",     # Microsoft.Web/sites
         "keyVaultSecretId", # Microsoft.Network/applicationGateways sslCertificates - this is actually a uri created with reference() and concat /secrets/secretname
-        "keyId", # Microsoft.Cdn/profiles urlSigningKeys
-        "subscriptionId", # Microsoft.Cdn/profiles urlSigningKeys
-        "StartingDeviceID" # SQLIaasVMExtension > settings/ServerConfigurationsManagementSettings/SQLStorageUpdateSettings
+        "keyId",            # Microsoft.Cdn/profiles urlSigningKeys
+        "subscriptionId",   # Microsoft.Cdn/profiles urlSigningKeys
+        "StartingDeviceID"  # SQLIaasVMExtension > settings/ServerConfigurationsManagementSettings/SQLStorageUpdateSettings
     )
 
     if ($exceptions -contains $myIdFieldName) { # We're checking resource ids, not tenant IDs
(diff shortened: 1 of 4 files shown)


bmoore-msft:
can you expand a bit on what you're trying to flag with the regex?

Also, perhaps a separate issue, it seems like this has the potential to exclude too many resource - i.e. makes sense for a /workflow, but this will also exclude any other resource with a definition property.  TBH I don't know of any, but if there are now (or in the future) this would except them.

Is it possible to add to the exceptions list or is the list of things that could be in a workflow not finite enough?

bobbyangers:
Hi Brian.
The issue I am running quite a bit since we are using Logic App extensively.      Thanks for considering my PR.

I am definitely open to suggestions or a better way, than having to "suppress" the check altogether... since it would apply to all resource in my folders, not just logic apps.

To my knowledge, no other resource use the **definition** as a root node.  So my intent was to skipping  the validation check against the "id" for everything inside the "workflow" section.  The logic app also has a different syntax for retrieving parameters,

Some examples:

    This should not fail cause the tests to fail, there it`s part of the workflow defintition,  
` "triggers": {

      "manual": {

        "correlation": {

        "clientTrackingId": "@{coalesce(triggerOutputs().headers?['x-email-correlation-id'], guid())}"
   },
 },`
 or this
`
 "inputs": {

      "value": {

        "correlationId": "@{coalesce(triggerOutputs().headers?['x-email-correlation-id'], guid())}",

       "emailId": "@body('Get_email_(V2)_-_Processing')?['id']"
    }
}`

bmoore-msft:
I don't know of any other resource that has the property either - the risk is that if one comes a long it would result in a "false negative" in the test...  So that's why I was thinking of more robustness there.  I don't think the risk is high, but as I say that it strikes me as a comment that will someday be used in a meme.

So it sounds like the list of json properties that could end in "id" are not finite, so the exception list is not a good plan?

Rather than except the property, should we consider allowing the @ syntax as propert syntax?  That wouldn't work if literal values are expected there, e.g.
```
"manual": {
  "emailId": "admin@contoso.com"
}
```

bobbyangers:
Hi again.

In regards to the exception list, is not finite; and everyone would have his own list; and that doesn't seem like the best option.  I think that the list of exceptions should only be items that are part of the schema; not user defined.  Like in this case, **emailId** or **clientTrackingId** ,  contextual to the logic app, not ARM.

Allowing the **@** syntax might be an option.  

So you would **-not** match the current [ ] syntax **AND** **-not** match `@ { ... }`  **AND** **-not** match `@function() ` 

I can give it a try.  

Having this said, since this should only apply "within" a workflow, can you think of anyway to identify if the current node is in that context ?

arm-ttk/testcases/deploymentTemplate/IDs-Should-Be-Derived-From-ResourceIDs.test.ps1
```diff
@@ -96,12 +120,11 @@ foreach ($id in $ids) { # Then loop over each object with an ID
     # - single quote on parameters and variables (resourceId first parameters may not be a literal string)
     #
     $exprMatch = "\s{0,}\[\s{0,}($($allowedExpressions -join '|' ))\s{0,}\(\s{0,}"
-
-    if ($expandedId -is [string] -and ` #if it happens to be an object property, skip it
```
bobbyangers, line 100: This was moved before the other check

unit-tests/IDs-Should-Be-Derived-From-ResourceIDs/Pass/exceptions-that-should-pass.json
```diff
@@ -30,6 +30,24 @@
...
+                "triggers": {
+                   "manual": {
+                      "correlation": {
+                         "clientTrackingId": "@{coalesce(triggerOutputs().headers?['id'], guid())}"
+                      },
+                      "emailId": "@body('Get_email_(V2)_-_Processing')?['id']"
+                   }
+                }
+            }
+          }
```
bobbyangers, line 50: I don't know if this is too specific or not Please advise

arm-ttk/testcases/deploymentTemplate/IDs-Should-Be-Derived-From-ResourceIDs.test.ps1
```diff
@@ -96,12 +120,11 @@ foreach ($id in $ids) { # Then loop over each object with an ID
...
-        $expandedId -notmatch $exprMatch  ){
+    if ($expandedId -notmatch $exprMatch  ){
             Write-Error "Property: `"$($id.propertyName)`" must use one of the following expressions for an resourceId property:
-            $($allowedExpressions -join ',')" `
-             -TargetObject $id -ErrorId ResourceId.Should.Contain.Proper.Expression
+            $($allowedExpressions -join ',')
+            Path: [$($id.JsonPath)]"`
+            -TargetObject $id -ErrorId ResourceId.Should.Contain.Proper.Expression
```
bobbyangers, line 129: Added to the output to give a little more context

bobbyangers (commented):
Please review my last commits

unit-tests/IDs-Should-Be-Derived-From-ResourceIDs/Pass/exceptions-that-should-pass.json
bmoore-msft: Looks ok, can you add a few more test cases?

PASS
- if appropriate, whitespace in and around the @ - e.g. for arm template expressions, something like " [ concat ( 'test' )]" is valid in ARM, if it is also in logic apps add some cases for that.

FAIL
- Some failures to make sure we're not too permissive, e.g. "[concat('@{test}'}]"

bobbyangers:
Ok,  for the tests.

I’ll double check about the spaces.

As for the fail... the regex “starts with @“ so ok for that too.

Will do that ASAP.

--Robert Angers
m:514-557-8074

Sent from my iPhone

On Oct 26, 2020, at 11:54, Brian Moore <notifications@github.com> wrote:

﻿

@bmoore-msft commented on this pull request.

________________________________

In unit-tests/IDs-Should-Be-Derived-From-ResourceIDs/Pass/exceptions-that-should-pass.json<https://eur05.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FAzure%2Farm-ttk%2Fpull%2F218%23discussion_r512070972&data=04%7C01%7C%7C1c41acc66a9b4502373d08d879c7728c%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C637393244805207532%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C1000&sdata=44g9f%2Fot%2FmER2pMkdYqzMqyrb4HfuIjguwUje6TOaIc%3D&reserved=0>:

> +        {
+          "type": "Microsoft.Logic/workflows",
+          "apiVersion": "2019-05-01",
+          "properties": {
+            "definition": {
+                "$schema": "https://schema.management.azure.com/providers/Microsoft.Logic/schemas/2016-06-01/workflowdefinition.json#",
+                "contentVersion": "1.0.0.0",
+                "triggers": {
+                   "manual": {
+                      "correlation": {
+                         "clientTrackingId": "@{coalesce(triggerOutputs().headers?['id'], guid())}"
+                      },
+                      "emailId": "@body('Get_email_(V2)_-_Processing')?['id']"
+                   }
+                }
+            }
+          }


Looks ok, can you add a few more test cases?

PASS

  *   if appropriate, whitespace in and around the @ - e.g. for arm template expressions, something like " [ concat ( 'test' )]" is valid in ARM, if it is also in logic apps add some cases for that.

FAIL

  *   Some failures to make sure we're not too permissive, e.g. "[concat('@{test}'}]"

unit-tests/IDs-Should-Be-Derived-From-ResourceIDs/Pass/logicapp-id-that-should-pass.json
```diff
@@ -0,0 +1,25 @@
+﻿{
```
bmoore-msft, line 1: ok, you lost me... what's supposed to fail here?

bmoore-msft (changes requested):

unit-tests/IDs-Should-Be-Derived-From-ResourceIDs/Pass/logicapp-id-that-should-pass.json
bobbyangers: Nothing should fail... it's the logicapp-id-**that-should-pass**.json

arm-ttk/testcases/deploymentTemplate/IDs-Should-Be-Derived-From-ResourceIDs.test.ps1
```diff
@@ -56,16 +56,42 @@ foreach ($id in $ids) { # Then loop over each object with an ID
...
         Write-Error "Blank ID Property found: $($id | Out-String)" -TargetObject $id -ErrorId ResourceId.Is.Missing
         continue
     }
     $expandedId = Expand-AzTemplate -Expression $myId -InputObject $TemplateObject -Exclude Parameters # then expand it.
+
+    if ($expandedId -isnot [string]) { #if it happens to be an object property, skip it
+        continue
+    };
+
+     # Check that it uses one of the allowed expressions - can remove variables once Expand-Template does full eval of nested vars
```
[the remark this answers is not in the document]
bobbyangers: the rules are well explained; maybe later we can put even more specific rules.

unit-tests/IDs-Should-Be-Derived-From-ResourceIDs/Pass/logicapp-id-that-should-pass.json
bmoore-msft: Sorry I meant this for the failure case...

unit-tests/IDs-Should-Be-Derived-From-ResourceIDs/Fail/logicapp-fail.json
```diff
@@ -0,0 +1,25 @@
+﻿{
+    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
+    "resources": [
+        {
+          "type": "Microsoft.Logic/workflows",
+          "apiVersion": "2019-05-01",
+          "name": "mylogicApp",
+          "properties": {
```
bmoore-msft, line 8: what is meant to fail in this case?

bobbyangers: I put that as an example....
    #  Valid expression examples
    #    1  =>   @{coalesce(triggerOutputs().headers?['id'], guid())}
    #    2  =>   @body('Get_email_(V2)_-_Processing')?['id']

And also put a reference to the doc in the comments: 
https://docs.microsoft.com/en-us/azure/logic-apps/workflow-definition-language-functions-reference

What would you want more?

unit-tests/IDs-Should-Be-Derived-From-ResourceIDs/Fail/logicapp-fail.json
```diff
@@ -0,0 +1,25 @@
...
+            "definition": {
+                "$schema": "https://schema.management.azure.com/providers/Microsoft.Logic/schemas/2016-06-01/workflowdefinition.json#",
+                "contentVersion": "1.0.0.0",
+                "triggers": {
+                   "manual": {
+                      "correlation": {
+                         "clientTrackingId": "@guid()"
+                      },
+                      "emailId": "somebody@somwhere.com"
```
bobbyangers, line 17: this is an email; not an email Id and is not a valid function reference

unit-tests/IDs-Should-Be-Derived-From-ResourceIDs/Fail/logicapp-fail.json
```diff
@@ -0,0 +1,25 @@
...
+                   },
+                   "somedataId": "data"
```
bobbyangers, line 19: this should fail, just a simple string

bobbyangers (commented):
Thanks for your question.

unit-tests/IDs-Should-Be-Derived-From-ResourceIDs/Pass/logicapp-id-that-should-pass.json
bobbyangers: Please see the explanations I put in the file and code.

bobbyangers:
I see you modidfied the same file.... Merge conflict ahead!!!

unit-tests/IDs-Should-Be-Derived-From-ResourceIDs/Fail/logicapp-fail.json
bmoore-msft: Ok, so this is interesting - should logicApp code follow the same pattern, e.g. generally, things that in in "id" should be an expression?

I guess that's already the case today, but if so, then we should probably restrict this test case to just the change we're making in the PR?

unit-tests/IDs-Should-Be-Derived-From-ResourceIDs/Fail/logicapp-fail.json
bmoore-msft: but this shouldn't fail should it?

bobbyangers:
Yes it should... doesn’t it.... I thought what is in that folder checks for fail?   Did I miss something?

--Robert Angers
m:514-557-8074

Sent from my iPhone

On Oct 29, 2020, at 12:01, Brian Moore <notifications@github.com> wrote:

﻿

@bmoore-msft commented on this pull request.

________________________________

In unit-tests/IDs-Should-Be-Derived-From-ResourceIDs/Fail/logicapp-fail.json<https://eur04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FAzure%2Farm-ttk%2Fpull%2F218%23discussion_r514375835&data=04%7C01%7C%7C9dda34e6060f401437f208d87c23d4d1%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C637395840614407657%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C1000&sdata=Ri3jbYOPSnXRoJkE9jIuQTXJFy5273kIzfwMY8VwFo8%3D&reserved=0>:

> +    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
+    "resources": [
+        {
+          "type": "Microsoft.Logic/workflows",
+          "apiVersion": "2019-05-01",
+          "name": "mylogicApp",
+          "properties": {
+            "definition": {
+                "$schema": "https://schema.management.azure.com/providers/Microsoft.Logic/schemas/2016-06-01/workflowdefinition.json#",
+                "contentVersion": "1.0.0.0",
+                "triggers": {
+                   "manual": {
+                      "correlation": {
+                         "clientTrackingId": "@guid()"
+                      },
+                      "emailId": "somebody@somwhere.com"


but this shouldn't fail should it?

bmoore-msft:
I don't think this:

```"emailId": "somebody@somwhere.com"```

should fail for a logic app.  What I'm thinking is that we're trying to mix best practices for an arm template and best practices for the logicApp language.  The rules are not likely to be the same.

bobbyangers:
Hi @bmoore-msft 

> The rules are not likely to be the same.

I do agree; we could assume that the Id should be a parameter or a formula;  I thought this was the intent of the validation rule.

I do also agree that the rules might not be the same;  Initially, I was checking that node was within a **definition** tag; but you asked me to change it to evaluation should include workflow **expressions**;

How can we know that the current node is "within" a workflow **definition**; 

So sorry if I misunderstood.  Please advise

bmoore-msft:
I think the challenge is that all rules are applied to all "json".  IOW, the rules we put in place for templates apply to logicApps and vice versa.  I think the "proper" solution it to understand the context when evaluating a rule, i.e. are we in a LogicApp or well, not.  And if so, apply the appropriate rules.

TBH I think the simple fix here might be simply skipping this test if we're inside a logicApp (or functionApp) resource - at least the "code" properties.

Thoughts?

bobbyangers:
I cannot say I disagree... that was my initial commit suggestion.

`
if ($id.JsonPath -match '\.properties(\[\d{1,2}\])?\.definition(\[\d{1,2}\])?\.')
   continue;
`
       
An you thought at the time that, it was too "dangerous" to assume **definition** was only used.   To be quite franc;   I have not seen anywhere else; the the api is not likely to add it somewhere else.

It is true that the data contract is from 2016, but has not changed since.

Anyway;  I would be happy to but it back the way it was originally intended.

bmoore-msft:
Yeah, that's close I just want to make sure we're future proofed...  So if we had a code path that throws out the definition property (or kicks in a different check) if and only if this is a logicApp resource, then I think we found the spot.

Not sure how exactly to do that, though there are a couple patterns (apiVersion test looks at resourceType)...

@StartAutomating may have some other ideas on what we could leverage if you're still up for it...

bobbyangers:
Ok great, I’ll take a look at those examples and come up with maybe a new PR, you then reject and close this one then.

--Robert Angers
m:514-557-8074

Sent from my iPhone

On Nov 5, 2020, at 16:57, Brian Moore <notifications@github.com> wrote:

﻿

Yeah, that's close I just want to make sure we're future proofed... So if we had a code path that throws out the definition property (or kicks in a different check) if and only if this is a logicApp resource, then I think we found the spot.

Not sure how exactly to do that, though there are a couple patterns (apiVersion test looks at resourceType)...

@StartAutomating<https://nam12.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FStartAutomating&data=04%7C01%7C%7Cd5e4cf7b31d946a2c1e508d881d5c08c%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C637402102334344849%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C1000&sdata=yb7ksLih84icHjbaEuHf6qQdj7KbJror2TahCQ%2FGN4E%3D&reserved=0> may have some other ideas on what we could leverage if you're still up for it...

bobbyangers:
here is another example of false negative....

									"inputs": {
										"body": {
											"id": "@items('For_each')?['id']"
										},
										"host": {
											"connection": {
												"name": "@parameters('$connections')['cn-sql']['connectionId']"
											}
										},

PR #224, opened 2020-10-23
RegEx improvements

StartAutomating:
Structuring Regular Expressions used within arm-ttk


bmoore-msft:
Let's expand on the docs and make them more scenario focused - so when would I use one approach vs. the other and show an example of consuming in a test...

StartAutomating:
We would like some samples of simple patterns / generators inline with /Regex/README.md

arm-ttk/Regex/README.md
```diff
@@ -0,0 +1,49 @@
...
+*.regex.txt pattern files
+*.regex.ps1 pattern generators
+
+These regular expressions can be used within arm-ttk by calling a PowerShell ScriptBlock that matches the name of the directory/file.
+
+### Using Patterns
+
+
+.regex.txt files will contain a single Regular Expression.
+
```
bmoore-msft, line 21: Let's add an inline example for the content of the txt file in a code block

arm-ttk/Regex/README.md
```diff
@@ -0,0 +1,51 @@
...
+This expression can be used anywhere within arm-ttk.  
+
+For instance, using the pattern ARM_Expression, we can find all expressions within the text of a template:
```
bmoore-msft, line 24: add a step to talk about the hierarchy and how the name is derived, i.e. where does "ARM_Expression" come from

arm-ttk/Regex/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+~~~PowerShell
+$templateText | & ${?<ARM_Expression>}
```
bmoore-msft, line 27: add - this will find all the "matches" for the expression in templateText

arm-ttk/Regex/README.md
```diff
@@ -0,0 +1,51 @@
...
+~~~
+
+
+### Using Generators
+
+.regex.ps1 files will contain Regular Expression generators.
+
```
bmoore-msft, line 34: same as above lets put the the code inline

And we can also use a simpler example (doesn't have to be real) to highlight the use of a parameter to make the regex more flexible, e.g. whitespace+param+whitespace

arm-ttk/Regex/README.md
```diff
@@ -0,0 +1,51 @@
...
+These are PowerShell scripts that produce a regular expression based off of input.
+
+For instance, we can find all of the references to the VMSize parameter.
+
+~~~PowerShell
+$templateText | & ${?<ARM_Parameter>} -Parameter 'VMSize'
+~~~
+
+Generators need not require parameters.  To find all parameters referenced throughout the template, use:
```
bmoore-msft, line 43: you could use a generator to find all resources of a given type and return properties for that resource

arm-ttk/Regex/ARM/Expression.regex.txt
```diff
@@ -0,0 +1,14 @@
+﻿# Matches an Expression in an Azure Resource Manager Template
+\s{0,}                                                     # optional whitespace
```
bmoore-msft, line 2: can  we add some more detail up top about the depth and balancing groups piece (i.e. why is it relevant to the scenario)

arm-ttk/Regex/ARM/List_Function.regex.txt
```diff
@@ -0,0 +1,9 @@
+﻿(?>        # we don't want to flag a UDF that might be called "myListOfIps" so we need to check the char preceeding list*()
```
bmoore-msft, line 1: Or's should be more explicit in the comments

arm-ttk/Writing_TTK_Tests.md
```diff
@@ -0,0 +1,85 @@
...
+            # At this point, $foundQuote could contain either a double quote or a bracket
+            # It it was a double quote, the parameter reference is not in an expression.
+            if ($foundQuote.Value -eq '"') { 
+                # so we produce an error.
+                Write-Error -Message "Parameter reference is not contained within an expression: $($Parameter.Name)" -ErrorId Parameters.Must.Be.Referenced.In.Expression -TargetObject $parameter
+            }
+        }
+    }
+}
+~~~
```
bmoore-msft, line 85: Add a section or another file for how to supply unit tests

bmoore-msft (changes requested):
Remember to pull first since we made a couple changes during the review

arm-ttk/Regex/README.md
StartAutomating: Example / "Writing Generator" section added

arm-ttk/Regex/README.md
StartAutomating: We have determined we want a simpler example, using ARM_Parameter.

StartAutomating:
Upon discussion, we would also like to rename the following expressions:  ARM_Parameter -> ARM_Template_Parameter and ARM_Variable - > ARM_Template_Variable.

Any future expression whose name would be easily confused without clarifying should be named ARM_Template_*

arm-ttk/Regex/ARM/List_Function.regex.txt
```diff
@@ -1,9 +1,20 @@
...
+    \[     # A bracket (the function could be the first thing in the template expression)
+    |      # OR
+    \(     # An open parenthesis (the function may be contained in another function)
+    |      # OR
+    ,      # A comma (the function may be contained in another function and not the first argument in that function)
+    )      # and the (?>  ) syntax says this is not included in the match because we need to check for expressions explicitly below    
+    \s{0,} # We also don't need to include any preceeding whitespace in the match itself
+)
+(?<ListFunctionName>  # We want to capture the list function name
+list\w{1,}            # which will be list followed by any number of word characters
```
bmoore-msft, line 15: should we just do this with the TLE generator?

PR #229, opened 2020-11-10
The vlanId in Microsoft.Network/expressRouteCircuits is not an resourceId

asifma:
The vlanId in Microsoft.Network/expressRouteCircuits is not an resourceId...

More info here: https://docs.microsoft.com/en-us/azure/expressroute/expressroute-howto-routing-portal-resource-manager


asifma:
Hey, can this be approved and merged? :)

PR #231, opened 2020-11-13
nodes that contain id within logic app should be skipped

bobbyangers:


arm-ttk/testcases/deploymentTemplate/IDs-Should-Be-Derived-From-ResourceIDs.test.ps1
```diff
@@ -61,8 +61,10 @@ foreach ($id in $ids) { # Then loop over each object with an ID
         }
     }
 
-    
-    
+    # Skip this check if the property is within a logic app
+    if ( $id.ParentObject.type -imatch '^microsoft.logic/workflows$' ) {
```
bmoore-msft, line 65: I think the . may have some side effects in the regex but practically speaking if that was your type property the template has other problems.  I'm ok with this as is - @StartAutomating can you have a quick look?

bobbyangers:
I tried to do amend.... reopening

arm-ttk/testcases/deploymentTemplate/IDs-Should-Be-Derived-From-ResourceIDs.test.ps1
bobbyangers: Can you please reopen ?

bmoore-msft:
I can't reopen - probably because the branch is gone...  But you can just submit another.

Issue #232, opened 2020-11-17
@@odata.id causing IDs Should Be Derived From ResourceIDs


LeonB87:
I have the following code in my ARM-Template that does a HTTP call to the Graph API

```code
{
	"HTTP_Add_Groupmember": {
		"runAfter": {
			"HTTP": [
				"Succeeded"
			]
		},
		"type": "Http",
		"inputs": {
			"authentication": {
				"audience": "https://graph.microsoft.com",
				"clientId": "@string(body('HTTP_Get_Client_Id').value)",
				"secret": "@string(body('HTTP_Get_Client_Secret').value)",
				"tenant": "@string(body('HTTP_Get_Tenant_Id').value)",
				"type": "ActiveDirectoryOAuth"
			},
			"body": {
				"@@odata.id": "https://graph.microsoft.com/v1.0/directoryObjects/@{json(decodeBase64(triggerBody()?['ContentData'])).userid}"
			},
			"headers": {
				"Content-type": "application/json"
			},
			"method": "POST",
			"uri": "@{concat('https://graph.microsoft.com/v1.0/groups/',string(body('HTTP').value[0].id),'/members/$ref')}"
		}
	}
}
```

This causes ARM-TTK to fail

![image](https://user-images.githubusercontent.com/10503724/99379782-67350e80-28c9-11eb-81a6-86b0b92636a9.png)

Adding "@@Odata.Id" to the exception list "fixes" this issue.

![image](https://user-images.githubusercontent.com/10503724/99379882-8633a080-28c9-11eb-9bb1-3a240f2d2996.png)

Creating Fork to suggest fix.

LeonB87:
Hi @bmoore-msft ; could you kindly look at the PR for this issue? Regards, Léon

LeonB87:
This is fixed in PR [#234 ](https://github.com/Azure/arm-ttk/pull/234 )

PR #233, opened 2020-11-17
Fix for issue #232 - Added @@odata.id as exception to IDs-Should-Be-Derived-From-ResourceIDs

LeonB87:
Fix for issue #232 - Added @@odata.id as exception to IDs-Should-Be-Derived-From-ResourceIDs

image

image


LeonB87:
fixed conflict in PR

bmoore-msft:
@LeonB87 - thanks for the ping...  I think this should be fixed by #234  essentially we want to disable this check in the logicApp workflows - that way we don't have to except each property (which I think is an infinite list for the workflow definition).

If it's not covered, can expand on the scenario a bit?

LeonB87:
@bmoore-msft You're right, that fix solved the issue (In a better way).

closing PR

Thanks!

PR #236, opened 2020-11-22

bobbyangers:


bobbyangers:
Hi @bmoore-msft ; thought it would be a good idea to test **parameters.json** files as well for correctness.  

At least these four rules,   schema, parameters exists,  parameter have value, contentversion is there and contains a verison

PR #237, opened 2020-11-24

bobbyangers:


unit-tests/DeploymentParameters-ContentVersion-Must-Exist/Fail/BadFormat.parameters.should.be.txt
```diff
@@ -0,0 +1 @@
+﻿ContentVersion.Bad.Format
```
bmoore-msft, line 1: the txt file is only needed if you want to refer to a file that's shared with other tests - so you should be able to remove all the txt files if that holds true for each of them...

unit-tests/DeploymentParameters-ContentVersion-Must-Exist/Pass/2019-04-01.parameters.json
```diff
@@ -0,0 +1,4 @@
+﻿{
+    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentParameters.json#",
+    "contentVersion": "1.0.0.0"
```
bmoore-msft, line 3: can we vary this a bit to test the regex some (though if we change it to match the schema probably less interesting)

unit-tests/DeploymentParameters-Parameters-Property-Must-Exist/Fail/NotExist.parameters.should.be.ps1
```diff
@@ -0,0 +1,17 @@
+﻿param(
```
bmoore-msft, line 1: do we need this file (in the unit test folder)?  Not sure what it does...

unit-tests/DeploymentParameters-Parameters-Property-Must-Exist/Pass/HasParameters.parameters.json
```diff
@@ -0,0 +1,4 @@
+﻿{
+    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentParameters.json#",
+    "parameters": {}
```
bmoore-msft, line 3: add contentVersion to this and you can reuse the file in other tests

arm-ttk/testcases/deploymentParameters/DeploymentParameters-Parameters-Property-Must-Have-Value.test.ps1
```diff
@@ -0,0 +1,25 @@
+<#
+.Synopsis
+    Ensures that parameters have a value
+.Description
+    Ensures that all parameters have a property 'value'
```
bmoore-msft, line 5: this needs to account for reference parameters...

https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/key-vault-parameter?tabs=azure-cli#reference-secrets-with-static-id

Probably just a simple "or" case...

unit-tests/DeploymentParameters-Parameters-Property-Must-Have-Value/Pass/HasValid.parameters.json
```diff
@@ -0,0 +1,8 @@
+﻿{
+    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentParameters.json#",
+    "parameters": {
+        "someKey" : {
+            "value" : "some data"
```
bmoore-msft, line 5: should add other param types here too:
- object, array, bool, int
- reference params

Since they are all passing you could put them into the same file

unit-tests/DeploymentParameters-Schema-Is-Correct/Fail/BadSchema.parameters.should.be.ps1
```diff
@@ -0,0 +1,17 @@
+﻿param(
```
bmoore-msft, line 1: as above, not sure what this file is for...

bmoore-msft (changes requested):
One thing we should pause on this is that the param file doesn't have to have this schema... it can be just the body of the parameters property or it can be the full schema.  Both are valid for ARM to consume.  I assume in that case we'd want to skip all the tests?  or we need to modify how we identify the files and run the tests...

We should doc that case specifically.

arm-ttk/Test-AzTemplate.ps1
```diff
@@ -305,6 +309,12 @@ Each test script has access to a set of well-known variables:
                 }
 
                 if (-not $matchingGroups) { continue }
+                if ($fileInfo.Schema -like '*deploymentParameters*') {
+                    $isMainTemplate = 'mainTemplate.parameters.json', 'azureDeploy.parameters.json', 'prereq.azuredeploy.parameters.json' -contains $fileInfo.Name
```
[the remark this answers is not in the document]
bobbyangers: @bmoore-msft  do we also modify it for the "azureDeploy.json" (line 319) ?   I was just following the set values for the other block.

bobbyangers:
> One thing we should pause on this is that the param file doesn't have to have this schema... it can be just the body of the parameters property or it can be the full schema. Both are valid for ARM to consume. I assume in that case we'd want to skip all the tests? or we need to modify how we identify the files and run the tests...
> 
> We should doc that case specifically.

Because of the way the files are selected; the validation would be skipped all together, so no harm done, no ?

Since ARM does consumes it anyway, then maybe; would be a good idea to issue a warning that if the file has the pattern ** 
 *.parameters.json** and no schema; then issue a warning; as it's good practice to put one.

arm-ttk/testcases/deploymentParameters/DeploymentParameters-ContentVersion-Must-Exist.test.ps1
```diff
@@ -0,0 +1,29 @@
...
+    Write-Error -ErrorId Parameters.Missing.ContentVersion -Message "contentVersion property must exist in the template"
+    return;
+} 
+
+if ($TemplateObject.contentVersion -isnot [string]) {
+    Write-Error -ErrorId ContentVersion.Not.String -Message "contentVersion must be string" -TargetObject $TemplateObject.contentVersion 
+    return;
+} 
+
+if ($TemplateObject.contentVersion -notmatch '\d{1,3}\.\d{1,3}\.\d{1,5}\.\d{1,5}') {
```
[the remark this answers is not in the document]
bobbyangers: [Understand the structure and syntax of ARM templates](https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/template-syntax)
> Version of the template (such as 1.0.0.0). You can provide any value for this element. Use this value to document significant changes in your template. When deploying resources using the template, this value can be used to make sure that the right template is being used.

So using a numeric version, recommended, but not restricted;   should we just issue a warning then ?

arm-ttk/Test-AzTemplate.ps1
```diff
@@ -288,6 +288,10 @@ Each test script has access to a set of well-known variables:
                                 if ($fn -match $key) { # check to see if it's name matches the key
                                     $key; continue # (this handles CreateUIDefinition.json, even if no schema is present).
                                 }
+                                if ($key -eq 'DeploymentParameters' -and # Otherwise, if we're checking the deploymentTemplate
```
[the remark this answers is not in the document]
bobbyangers: That is a good question.   I have seen some people use ***.parameters.dev.json**, ***.parameters.sbx.json** (sandbox) ;   so right now the way it was done for deploymentTemplate is to validate there was a **$schema** and go; that is my I chose to use the same technic.

As always; I am open to suggestions.

unit-tests/DeploymentParameters-Parameters-Property-Must-Exist/Fail/NotExist.parameters.should.be.ps1
bobbyangers: I consulted the other tests, they use that file to validate the proper error message.  I tried to follow that pattern as well.  If you think just validating the error code; that is fine with me.

unit-tests/DeploymentParameters-Parameters-Property-Must-Exist/Pass/HasParameters.parameters.json
bobbyangers: done;

As for the reuse; you'll have to show me how (or point me to anther example).

unit-tests/DeploymentParameters-Parameters-Property-Must-Have-Value/Pass/HasValid.parameters.json
bobbyangers: done.

arm-ttk/testcases/deploymentParameters/DeploymentParameters-Parameters-Property-Must-Have-Value.test.ps1
bobbyangers: Great catch.... In fact I did use that option before.
I'll do that.
Actually added a little more....

unit-tests/DeploymentParameters-Schema-Is-Correct/Fail/BadSchema.parameters.should.be.ps1
bobbyangers: removed

bmoore-msft (changes requested):
There were one or two unresolved comments from the previous review...  other notes:

- to reuse a test file you can follow this pattern https://github.com/Azure/arm-ttk/blob/master/unit-tests/Location-Should-Not-Be-Hardcoded/Pass/MarketplaceSample.path.txt - basically a file with *.path.txt that contains a path to the file to use - thinking about it, your files are trivial enough it may not matter so I'll leave that up to you...

- the extraneous txt files in the unit-test folders should also be removed - I think that's related to the PS1 files.  I think those may have been missed in PR.  I'm going to file an issue to remind us...

arm-ttk/testcases/deploymentParameters/DeploymentParameters-Parameters-Property-Must-Have-Value.test.ps1
```diff
@@ -5,21 +5,55 @@
...
-    # check if the property exist on the hashtable Value property for the key 'value'
-    if ( -not $p.Value.PsObject.Properties.match('value').Count ) {
-        Write-Error -ErrorId Parameters.Parameter.Missing.Value -Message "'$($p.Name)' must have a property 'value'" -TargetObject $p.Value 
+  # check if the property exist on the hashtable Value property for the key 'value'
+  if ( -not ($p.Value.PsObject.Properties.match('value').Count -or $p.Value.PsObject.Properties.match('reference').Count )) {
+    Write-Error -ErrorId Parameters.Parameter.Missing.Value -Message "'$($p.Name)' must have a property 'value' or 'reference'" -TargetObject $p.Value 
+    continue
+  }
+
+  if ( $p.Value.PsObject.Properties.match('value').Count){
```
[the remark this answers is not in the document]
bobbyangers: Hi,   these are new tests that were added on top of checking just for "value", we now check for 'reference' as well, and make sure it is well formed.
Now this formula, is meant to check the presence for the 'value' in the properties collection.  I tried several options, and this is the one I found was working.
I tried => 'value' in $p.Value.PsObject.Properties.Name but could not get it to pass the tests.

unit-tests/arm-ttk.test.functions.ps1
```diff
@@ -111,25 +111,28 @@ function Test-TTKFail {
             $ttkResults = Get-Item -Path $testFile.Fullname | 
                 Test-AzTemplate @ttkParams 
             if (-not $ttkResults) { throw "No Test Results" }
-            if (-not ($ttkResults | Where-Object {$_.Errors })) {
-                throw 'Errors were expected'
-            }
             if (Test-Path $targetTextPath) { # If we have a .should.be.txt
                 $targetText = [IO.File]::ReadAllText($targetTextPath).Trim() # read it
+                if (-not ($ttkResults | Where-Object {$_.Errors })) {
```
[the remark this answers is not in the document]
bobbyangers: As I as was testing, and using the .txt files I was getting simply getting "expected errors"... I thought it be better to know which error it was expecting; So by checking it after looking at the txt file, then I could append the content and obtain a more complete error message.

It did however force me to repeat in the "else";   but I thought that the extra info was worth it.

bobbyangers:
> There were one or two unresolved comments from the previous review... other notes:
> 
> * to reuse a test file you can follow this pattern https://github.com/Azure/arm-ttk/blob/master/unit-tests/Location-Should-Not-Be-Hardcoded/Pass/MarketplaceSample.path.txt - basically a file with *.path.txt that contains a path to the file to use - thinking about it, your files are trivial enough it may not matter so I'll leave that up to you...

I'll look at that.

> * the extraneous txt files in the unit-test folders should also be removed - I think that's related to the PS1 files.  I think those may have been missed in PR.  I'm going to file an issue to remind us...

I thought the request was to remove the PS1 and keep the TXT.. did I get it wrong ? In any case, either is fine with me, but I think it's a good idea to test for the specific error.   I used other tests to discover that,
After some digging both can be used;  the TXT can be the ErrorId or some text to be searched in "any" of the error messages generated.
the PS1; is simply executed and expected to throw if the expected conditions are not met.

unit-tests/DeploymentParameters-ContentVersion-Must-Exist/Pass/2019-04-01.parameters.json
bobbyangers: Sorry, I am not sure I follow...  
I modified the check where to check the version string and issue a warning;   the spec says we can put anything, but it's recommended to use "version" strings.

bobbyangers:
Reshaped the commit (squashed them) to simplify

PR #240, opened 2020-11-26

bobbyangers:

For the full conversation:  please see [PR-237](https://github.com/Azure/arm-ttk/pull/237)


bobbyangers:
Hi @bmoore-msf, hope all is well in these crazy time!

Did you have a chance to look at this updated PR regarding Logic Apps?

arm-ttk/Test-AzTemplate.ps1
```diff
@@ -305,8 +309,14 @@ Each test script has access to a set of well-known variables:
                 }
 
                 if (-not $matchingGroups) { continue }
+                if ($fileInfo.Schema -like '*deploymentParameters*') {
+                    $isMainTemplate = 'maintemplate.parameters.json', 'azuredeploy.parameters.json', 'prereq.azuredeploy.parameters.json' -contains $fileInfo.Name
+                    $templateFileName = $fileInfo.Name
```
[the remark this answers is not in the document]
bobbyangers: Ok, I was just following the current method...

I am not sure how to go about doing what you asked.

unit-tests/DeploymentParameters-Parameters-Property-Must-Have-Value/Fail/BadSecretNameKeyVault.parameters.json
```diff
@@ -0,0 +1,13 @@
+﻿{
+    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentParameters.json#",
+    "parameters":{
+      "someKeyVaultKey": {
+        "reference": {
+          "keyVault": {
+            "id" :"/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/rg1/providers/Microsoft.KeyVault/vaults/myvault"
```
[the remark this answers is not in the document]
bobbyangers: In the regex, i used -imatch which already does a case insensitive match....

So what are you suggesting?

unit-tests/DeploymentParameters-Parameters-Property-Must-Have-Value/Pass/HasValid.parameters.json
```diff
@@ -0,0 +1,29 @@
...
+        "value" : {}
+      },
+      "someBool" : {
+        "value" : true
+      },
+      "someInt" : {
+        "value" : 1
+      },
+      "someReference" : {
+        "value" : "[reference('other').output.value]"
```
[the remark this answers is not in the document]
bobbyangers: Ok, I'll add a test for that in the fail section.

bobbyangers:
@bmoore-msft  any chance you'll be looking at this soon ?

bobbyangers:
Hi @bmoore-msft,
Any chance these get approuve soon?

bobbyangers:
Hi @bmoore-msft ,   I am not sure what do to next.  Can you please point me in the right direction ?

bobbyangers (commented):
done

bobbyangers:
hi @bmoore-msft,  
Happy New Year.  Hope all is well.
how's it going ?   Still waiting for your input to close this....

rangers-goodreason:
Hi @bmoore-msft , I have taken some time to rebase this PR;  

please let me know if there is anything missing

bobbyangers:
Hi @bmoore-msft , I have taken some time to rebase this PR;

please let me know if there is anything missing

bobbyangers:
Hi @bmoore-msft  where do we stand on this ?

unit-tests/DeploymentParameters-Parameters-Property-Must-Have-Value/Fail/BadSecretNameKeyVault.parameters.json
```diff
@@ -0,0 +1,13 @@
...
+          },
+          "secretName": "some_invalid_value"
```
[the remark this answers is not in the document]
bobbyangers: no longer applicable

unit-tests/DeploymentParameters-Parameters-Property-Must-Have-Value/Pass/HasValid.parameters.json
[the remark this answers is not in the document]
bobbyangers: no longer applicable

arm-ttk/Test-AzTemplate.ps1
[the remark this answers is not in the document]
bmoore-msft: Ok, sorry for the delay here...  I think we want to have a new set of "global" vars for the param file equivalent, e.g.
```
$parameterFileName
$parameterObject
$parameterText
```
Rather than overloading the $template* variables.  Right now I can't think of a problem it will cause, but if we do find one later, cleaning up will be much harder.  So better to separate them from the start.

arm-ttk/Test-AzTemplate.ps1
```diff
@@ -311,6 +314,12 @@ Each test script has access to a set of well-known variables:
                 }
 
                 if (-not $matchingGroups) { continue }
+                if ($fileInfo.Schema -like '*deploymentParameters*') {
+                    $isMainTemplate = 'maintemplate.parameters.json', 'azuredeploy.parameters.json', 'prereq.azuredeploy.parameters.json' -contains $fileInfo.Name
```
bmoore-msft, line 318: change this to $isMainTemplateParameters

arm-ttk/Test-AzTemplate.ps1
```diff
@@ -293,9 +293,12 @@ Each test script has access to a set of well-known variables:
                                 if ($fn -match $key) { # check to see if it's name matches the key
                                     $key; continue # (this handles CreateUIDefinition.json, even if no schema is present).
                                 }
+                                if ($key -eq 'DeploymentParameters' -and # Otherwise, if we're checking the deploymentTemplate
+                                'maintemplate.parameters.json', 'azuredeploy.parameters.json', 'prereq.azuredeploy.parameters.json' -contains $fn) { # and the file name is something we _know_ will be an ARM template
```
bmoore-msft, line 297: should this just be: *.parameters.json ?

Also what happens if I don't name my param file with this pattern (or have no param file) - are the tests simply not executed?  I think so, but want to confirm.

arm-ttk/testcases/deploymentParameters/DeploymentParameters-Parameters-Property-Must-Have-Value.test.ps1
```diff
@@ -0,0 +1,65 @@
...
+.Synopsis
+    Ensures that parameters have a value
+.Description
+    Ensures that all parameters have a property 'value'
+#>
+param(
+  # The Template Object
+  [Parameter(Mandatory = $true, Position = 0)]
+  [PSObject]
+  $TemplateObject
```
bmoore-msft, line 11: this would need to change with the changes above, i.e. $parameterObject

arm-ttk/testcases/deploymentParameters/DeploymentParameters-Parameters-Property-Must-Have-Value.test.ps1
```diff
@@ -0,0 +1,65 @@
...
+
+Write-Debug $TemplateObject.parameters
+foreach ($p in $TemplateObject.parameters.psobject.properties) {
+    
+  # If the parameter name starts with two underscores,
+  if ($p.Name -like '__*') { continue } # skip it.
+
+  # check if the property exist on the hashtable Value property for the key 'value'
+  # just checking if the property is there might result in false negative, for example if the property contains:  $null, 0, False, empty string
+  if ( -not ($p.Value.PsObject.Properties.match('value').Count -or $p.Value.PsObject.Properties.match('reference').Count )) {
```
bmoore-msft, line 22: I think this can be simplified, e.g. first check to see if the required properties are present, must have 1 of the following:
```powershell
if($p.Value.PSObject.properties["value"] -or $p.Value.PSObject.properties["reference"]){
    continue
}
```

Then I think you can skip to [the current] line 36

arm-ttk/testcases/deploymentParameters/DeploymentParameters-Parameters-Property-Must-Have-Value.test.ps1
```diff
@@ -0,0 +1,65 @@
...
+      continue
+    }
+    else {
+      $kv = $kvRef.keyVault;
+      if(-not $kv.PsObject.Properties.match('id').Count){
+        Write-Error -ErrorId Parameters.Parameter.Missing.KeyVault.Id -Message "'$($p.Name)'.reference is missing have a property 'keyVault'" -TargetObject $p.Value.reference.keyVault
+        continue
+      }
+      # Resource Group naming restrictions, ref: https://github.com/toddkitta/azure-content/blob/master/articles/guidance/guidance-naming-conventions.md
+      # Keyvalut naming restrictions, ref: https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftkeyvault
```
bmoore-msft, line 49: typo

arm-ttk/testcases/deploymentParameters/DeploymentParameters-Parameters-Property-Must-Have-Value.test.ps1
```diff
@@ -0,0 +1,65 @@
...
+      elseif($kv.id -inotmatch '/subscriptions/([0-9a-f]{8}(-[0-9a-f]{4}){3}-[0-9a-f]{12})/resourceGroups/([\w\d_-]{1,64})/providers/Microsoft.KeyVault/vaults/([\w\d-]{3,24})') {
+        Write-Error -ErrorId Parameters.Parameter.Bad.KeyVault.Id -TargetObject $p.Value.reference.keyVault.id -Message "'$($p.Name)'.reference.keyValut.id contains improper value, `r`n Should be like : /subscriptions/<subscription id>/resourceGroups/<resource group name>/providers/Microsoft.KeyVault/vaults/<key vault name>"
```
bmoore-msft, line 51: typo in keyVault

unit-tests/DeploymentParameters-Parameters-Property-Must-Have-Value/Fail/BadSecretNameKeyVault.parameters.json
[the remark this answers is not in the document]
bmoore-msft: just having a test case that's checking that your -inotmatch doesn't change to -cnotmatch...

unit-tests/DeploymentParameters-Parameters-Property-Must-Have-Value/Pass/HasValid.parameters.json
```diff
@@ -0,0 +1,26 @@
...
+      "someBool" : {
+        "value" : true
+      },
+      "someInt" : {
+        "value" : 1
+      },
+      "someKeyVaultReference": {
+        "reference": {
+          "keyVault": {
+            "id": "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/rg1/providers/Microsoft.KeyVault/vaults/myvault"
```
bmoore-msft, line 20: add some alpha chars to the guid since your regex checks them

bmoore-msft (changes requested):
Sorry for the delay bobby, finally getting back to some of these...

I do like the stuff you're adding...

arm-ttk/testcases/deploymentParameters/DeploymentParameters-Parameters-Property-Must-Have-Value.test.ps1
bobbyangers: Hi @bmoore-msft ,

I am not sure about your claim.    the current code, splits the errors into more specific errors, 
1.  making sure either 'value' or 'reference'
2.  if value => cannot be a reference [somefunction()]
3.  if reference => checks the validity

as per my comment, not checking for Count can result in false positive when property contains a 'falsy' value.

It is important to allow, 0, false;   

I don't see how your proposal would cover those cases

arm-ttk/Test-AzTemplate.ps1
bobbyangers: Yes, as before, it did not check files it did that do not follow the pattern.

arm-ttk/testcases/deploymentParameters/DeploymentParameters-Parameters-Property-Must-Have-Value.test.ps1
bobbyangers: if we change this, then we would need to also change how this file gets called, the parameter is passed by name.

arm-ttk/Test-AzTemplate.ps1
bmoore-msft: For template files we check all json files and then have additional tests for the "main" templates...  seems like should follow that pattern here too, otherwise folks will expect thing are checked and they won't be...

arm-ttk/testcases/deploymentParameters/DeploymentParameters-Parameters-Property-Must-Have-Value.test.ps1
bmoore-msft: yes - we will...

As a side note, coming back to this PR I was still lost thinking we were looping on the template (not the params) so I definitely think we need to separate them...

arm-ttk/testcases/deploymentParameters/DeploymentParameters-Parameters-Property-Must-Have-Value.test.ps1
```diff
@@ -0,0 +1,65 @@
...
+#>
+param(
+  # The Template Object
+  [Parameter(Mandatory = $true, Position = 0)]
+  [PSObject]
+  $TemplateObject
+)
+
+Write-Debug $TemplateObject.parameters
+foreach ($p in $TemplateObject.parameters.psobject.properties) {
```
bmoore-msft, line 15: it's possible that the param file does not have a param object - a valid param file might just contain the name value pairs, e.g.

```json
{
  "adminUsername": {
    "value": "someuser"
  }
}
```

So we need to account for that case.

arm-ttk/testcases/deploymentParameters/DeploymentParameters-Parameters-Property-Must-Have-Value.test.ps1
bmoore-msft: The proposed would will tell you existence of the property, which I think is what's needed here.  If the property contains $null, 0, false, empty string - those are valid values for an arm template parameter.  So pass the existence and "is my value valid" test.

I think the part that jumping out is that the proposed sytnax is more readable than the ".match().count"

bmoore-msft (changes requested):

arm-ttk/Test-AzTemplate.ps1
bobbyangers: With James' help; we fixed this.

arm-ttk/testcases/deploymentParameters/DeploymentParameters-Parameters-Property-Must-Have-Value.test.ps1
bobbyangers: With James' help; done.

arm-ttk/testcases/deploymentParameters/DeploymentParameters-Parameters-Property-Must-Have-Value.test.ps1
bobbyangers: Is this an undocumented feature ?
According to this page;  
[Create Resource Manager parameter file](https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/parameter-files)

That does not seem to be an option.

Would it not be a case where a user might decide to not check a rule ?

arm-ttk/testcases/deploymentParameters/DeploymentParameters-Parameters-Property-Must-Have-Value.test.ps1
bmoore-msft: I don't see what you changed here?

bobbyangers: I will change it as you ask.

but what I understand from Powershell is that
if($p.Value.PSObject.properties["value"]) ==> false if the propertie is missing or 
contains a "falsy" value: 0, $null, false, empty string, empty array; 

if that is the case, then if the property exists, but contains a "falsy" value, then your suggestion will yield a false positive result.

arm-ttk/testcases/deploymentParameters/DeploymentParameters-Parameters-Property-Must-Have-Value.test.ps1
bmoore-msft: It probably is undocumented but valid so we need to figure out the difference and which tests to run...

bobbyangers: Hi @bmoore-msft, at the moment the are not being tested; so if the developer wanted to be tested, why not simply expect them to follow the documented features.

I understand what you want, however, pushing this update will enhance the product, and ignore some undocumented features; which by the way, is what it's doing now.

My vote is to push it as is; and enhance it later.

bmoore-msft: The main scenario is for generating params - a simple hashtable is easier to generate than ones with the extra properties/schema.  Though to be fair, those aren't often checked into source, so it just depends on when the ttk is executed.

If we push and find out we're breaking folks, we'd have to roll back.

@StartAutomating - any objections?

bobbyangers: That, or we could remove that particular test.

bobbyangers:
Hi @bmoore-msft ,

I have attempted to rebase, but it seems the tests are failing on this:
>> Resources-Should-Not-Be-Ambiguous.test
It looks like you just put an "exit" at the top of the file, so now the "should fail" are blocking this commit.

Describing Resources-Should-Not-Be-Ambiguous

  Context Fail
    [-] Validates ResourceTypeFirstParameter is flagged 21ms
      RuntimeException: Errors were expected
      at <ScriptBlock>, C:\Users\Robert\source\repos\arm-ttk\unit-tests\arm-ttk.test.functions.ps1: line 125
    [-] Validates trailing-slash is flagged 29ms
      RuntimeException: Errors were expected
      at <ScriptBlock>, C:\Users\Robert\source\repos\arm-ttk\unit-tests\arm-ttk.test.functions.ps1: line 125

  Context Pass
    [+] Validates NestedResourceType is correct 17ms
    [+] Validates SimpleResourceID is correct 18ms
    [+] Validates top-level-child-expression is correct 23ms
    [+] Validates topLevelChild is correct 19ms

bmoore-msft:
yep - moved those if you want to try again

arm-ttk/Test-AzTemplate.ps1
```diff
@@ -293,9 +293,12 @@ Each test script has access to a set of well-known variables:
                                 if ($fn -match $key) { # check to see if it's name matches the key
                                     $key; continue # (this handles CreateUIDefinition.json, even if no schema is present).
                                 }
+                                if ($key -eq 'DeploymentParameters' -and # Otherwise, if we're checking the deploymentTemplate
```
bmoore-msft, line 296: this seems like it's going to filter out tests, esp the schema test...  can we confirm the following.

- if the file is named like *.parameters.json AND
- the file has a property named "$schema"

run all of the parameter file tests.

The code here seems to filter on the schema property value which will mean that we skip schema test itself.

rangers-goodreason: Hi @bmoore-msft ,

> [290] foreach ($key in $TestGroup.Keys) { # If it didn't have a schema

I am not sure why you are asking to check for the schema ?

Plus I think this is getting checked at line 318

> [318]  if ($fileInfo.Schema -like '*deploymentParameters*')

bmoore-msft: parameter files have 2 [valid] schemas.  TLDR; one includes a $schema, contentVersion, and parameters property.  The other is a simple object with the contents of what would be in the parameters property in the first schema.

If we find a file named ```azuredeploy.parameters.json``` but it doesn't have a schema property, we should assume it won't have a parameters property either - and then skip the tests or adjust accordingly.  Adjusting is the right answer, but I think we can defer that to another PR.

This check:
```
if ($fileInfo.Schema -like 'deploymentParameters')
```

should be seconday, otherwise we will fail to test files named ```*.parameters.json``` but have the deploymentTemplate schema.

Essentially what this means, is that we will assert some file is a param file by only the filename (which I think is fine in this case).  Then we can test for more/all issues in the $schema property.

bobbyangers:
Hi Brian.

I am sorry, maybe we can organise a quick call with you or James?  I thought we would not support the scenario you are proposing;

--Robert Angers
m:514-557-8074

Sent from my iPhone

On Mar 12, 2021, at 13:48, Brian Moore ***@***.***> wrote:

﻿

@bmoore-msft commented on this pull request.

________________________________

In arm-ttk/Test-AzTemplate.ps1<https://na01.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FAzure%2Farm-ttk%2Fpull%2F240%23discussion_r593374925&data=04%7C01%7C%7Cca9037337c054de62ed108d8e58762a5%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C637511716917236711%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C1000&sdata=2nnMXfUxkCw90Z8G3OSBV6Uz31DnlapMzYAOg1xPrJU%3D&reserved=0>:

> @@ -293,9 +293,12 @@ Each test script has access to a set of well-known variables:
                                 if ($fn -match $key) { # check to see if it's name matches the key
                                     $key; continue # (this handles CreateUIDefinition.json, even if no schema is present).
                                 }
+                                if ($key -eq 'DeploymentParameters' -and # Otherwise, if we're checking the deploymentTemplate


parameter files have 2 [valid] schemas. TLDR; one includes a $schema, contentVersion, and parameters property. The other is a simple object with the contents of what would be in the parameters property in the first schema.

If we find a file named azuredeploy.parameters.json but it doesn't have a schema property, we should assume it won't have a parameters property either - and then skip the tests or adjust accordingly. Adjusting is the right answer, but I think we can defer that to another PR.

This check:

if ($fileInfo.Schema -like 'deploymentParameters')


should be seconday, otherwise we will fail to test files named *.parameters.json but have the deploymentTemplate schema.

Essentially what this means, is that we will assert some file is a param file by only the filename (which I think is fine in this case). Then we can test for more/all issues in the $schema property.

bobbyangers (commented):
I 'll wait for further instructions.

arm-ttk/Test-AzTemplate.ps1
bobbyangers: @bmoore-msft, @StartAutomating and myself reviewed the various points mentioned in the last commit,
1.   allowed ***.parameters.json** to be picked up
2. added guard clauses to account for $schemaless parameter files
3. renamed the tests files to better reflect intent

bobbyangers:
@bmoore-msft   Please latest updates done in concurence with @StartAutomating ;   to comply with your request regarding the **schemaless** scenario.

bmoore-msft:
Yea!

Issue #74, opened 2020-05-24
tags name containing id fails IDs Should Be Derived From ResourceIDs test


EmFl:
Hello,
I have a resource with a tag that is called projectid. This tag has nothing to do with azure IDs of course since it relates to our internal system for tracking projects.
If my understanding is correct I suppose a slight change is needed around here : 
https://github.com/Azure/arm-ttk/blob/9dec6bc4d16d1357a7dab5e64c51185ba1571e46/arm-ttk/testcases/deploymentTemplate/IDs-Should-Be-Derived-From-ResourceIDs.test.ps1#L42
To also ignore when path contains "tags".

Thanks

samrobillard:
Another case of the same issue is with policyDefinitionId. Should this property be added here?
https://github.com/Azure/arm-ttk/blob/d213230a40dab3a1b8c8447719be6b3b712b0bca/arm-ttk/testcases/deploymentTemplate/IDs-Should-Be-Derived-From-ResourceIDs.test.ps1#L65

bmoore-msft:
@samrobillard - we're working on a fix for the tags scenario, but policyDefinitionId should be a resourceId so wouldn't except that one... how are you seeing this manifest with policyDefinitionId?

bmoore-msft:
Moving this out of M5 per email conversation

bmoore-msft:
This should be fixed...

PR #258, opened 2020-12-16
Fixing #74 - Allowing ProjectID to not be a ResourceID.  

StartAutomating:
Fixing #74 - Allowing ProjectID to not be a ResourceID.  
Refactoring test to add -NotResourceIDs parameter if additional custom exemptions are needed.


bmoore-msft (changes requested):

bmoore-msft:
Going to close this as we want to 
1) leave this hard-coded instead of a param so that fixes make it back to this repo instead of commandlines all over the world
2) projectId is a tag we want to handle that scenario #274

Issue #37, opened 2020-04-15
adminUsername-Should-Not-Be-A-Literal does not catch hardcoded variables


bmoore-msft:
Taking a note while reviewing a PR... if the adminUsername property uses a variable and the variable is not an expression, this test won't find it... it looks like we are excluding top-level properties (technically second level properties) so we aren't finding vars with Find-JSONContent.

StartAutomating:
Just an FYI on this one:  because it uses the new Regex goodness, and the one expression that had an import issue, you'll need to approve #259 before this test will pass.

PR #260, opened 2020-12-16
Fixing #37 admin user name variable literal

StartAutomating:
Fixing #37


unit-tests/adminUsername-Should-Not-Be-A-Literal/Fail/Literal-Variable-Username.json
```diff
@@ -0,0 +1,22 @@
+{
+    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
+    "variables": {
+        "foo": {
+            "default": 1
```
bmoore-msft, line 5: 2 things:

Remove the object and default and use a literal spring, e.g.
"foo": "bar"

We also need to account for object variables:
"vmSettings": {
  "userName": "foo",
  "vmSize": "Standard_DS3"
}
Which would use:
"adminUsername": "[variables('vmSettings').userName]"

bmoore-msft (changes requested):

Issue #159, opened 2020-07-09
ensure resourceType in resourceId test does not end in a trailing /


bmoore-msft:
When checking resourceIds, make sure the type does not end with /...

Note that the ordinal position of the type param can vary within and across  resourceId functions...

Issue #140, opened 2020-07-09
resourceIds must contain the resourceGroup param when conditional or not in the template


bmoore-msft:
In any resourceId function, reference() function or list*() function

The *resourceId() functions must have at least one parameter that precedes the type parameter (type parameter can be identified as a string containing "/")
The list*() or reference() functions must have the full resourceId (i.e. cannot just use the resource name for the resourceId parameter)
in this case if the param/var is used as the name of a resource, and there is a condition on that resource, the test should throw

PR #264, opened 2020-12-17
ResourceID Ambiguity Tests.

StartAutomating:
Fixing #140 & Fixing #159 

Must accept PR for RegEx engine changes first.


arm-ttk/testcases/deploymentTemplate/Resources-Should-Not-Be-Ambiguous.test.ps1
```diff
@@ -0,0 +1,47 @@
...
+# The object representation of an Azure Resource Manager template.
+[Parameter(Mandatory,Position=0)]
+$TemplateObject,
+
+# The text of an Azure Resource Manager template
+[Parameter(Mandatory,Position=1)]
+$TemplateText
+)
+
+$resourceIdFunctions = $TemplateText | ?<ARM_Template_Function> -FunctionName resourceId
```
bmoore-msft, line 17: more detail on the test...

If the resource specified in the resourceId function (determined by type and name) is not defined in the template or is defined in the template with a condition property, then the resourceId function must have a parameter before the resourceType. e.g.

resourceId(parameters('storageResourceGroup'), 'Microsoft.Storage/storageAccounts', parameters('storageAccountName')

The first parameter can be anything as long as it's not a resourceType which we can assume is a literal string (even though technically it need not be).  We will not (and can document) that we are not flagging the use of resourceGroup().name here because we already have a test that does that (and don't want to duplicate the logic).

bmoore-msft (changes requested):
Need unit tests for:
- resource defined in template w/ trailing slash (and without)
- resource not defined in the template, must have function arg prior to the type
- resource defined in the template with the condition property (which is a peer of the type/name)

bmoore-msft:
From the conversation... we need to match the resourceId in the function with a resource in the template and may need additional parsing in the engine.  Scenarios

```
parameter
    storageC (some existing storage account name passed in to the template)

resources
    storageA
    storageB+condtion
    ...
    
    resourceId(Microsoft.Storage/storageAccounts, parameter(storageC)) - error because not defined in the template and no rgName param
    resourceId(storageC-rgName, Microsoft.Storage/storageAccounts, parameter(storageC)) - correct

    resourceId(Microsoft.Storage/storageAccounts, parameter(storageB)) - error because there is a condtion on B (and no rgName)

    resourceId(Microsoft.Storage/storageAccounts/blob, parameter(storageA), blobname) - no error, defined in the template and no condition


/subs/rg/{name}Microsoft.Storage/storageAccounts/parameters(storageA)/blob/blobname


type: Microsoft.Storage/storageAccounts/blob
name: concat(parameter(A), /, blobName)
name: literal/blob
name: format({0}/{1}, paramter(A), blobName)

Could also be a nested child resource
type: blob
name: blobname
```

We won't worry about the scope property for now since it's only used on extension resources atm

arm-ttk/testcases/deploymentTemplate/Resources-Should-Not-Be-Ambiguous.test.ps1
```diff
@@ -14,34 +22,69 @@ $TemplateObject,
...
     $foundApiVersion = ''
-    for ($n = 0 ;$n -lt $resourceIdParameters.Count;$n++) {
+    $additionalParameters = @(for ($n = 0 ;$n -lt $resourceIdParameters.Count;$n++) {
         if ($resourceIdParameters[$n] -like '*/*') {
-            $foundResourceType = $resourceIdParameters[$n]
-                    
-            if ($n -eq 0) {
-                Write-Error "At least one parameter must preceed the resource type" -TargetObject $rid -ErrorId 'ResourceID.Missing.Name'
-                continue nextResourceId
+            $foundResourceType = $resourceIdParameters[$n]
```
bmoore-msft, line 36: improve the "if" guard to make sure we stop at the first match of *.*/* and not go beyond (even though that's a funky scenario)

If we do not find a match of *.*/*, then throw an error , that means we might have an RP that doesn't have a period in the namespace

arm-ttk/testcases/deploymentTemplate/Resources-Should-Not-Be-Ambiguous.test.ps1
```diff
@@ -14,34 +22,69 @@ $TemplateObject,
...
 
         if ($foundResourceType -match "/'\s{0,}$") {
             Write-Error "ResourceType has a trailing slash '$foundResourceType'" -TargetObject $rid -ErrorId 'ResourceID.Trailing.Slash'
-        }
+        }        
+
+        # Find all resources of this type within the template
+        $resourcesOfType = 
+            @(Find-JsonContent -InputObject $TemplateObject.resources -Key type -Value $resourceTypeName) +
+            @(Find-JsonContent -InputObject $TemplateObject.resources -Key type -Value @($resourceTypeName -split '/' -ne '')[-1])
```
bmoore-msft, line 67: make sure we can get the different variations of nested child resources...  walk "up" until we have the entire resourceType.

type: blobservices
type: blobservices/containers
type: containers

Microsoft.Storage/storageAccounts/blobservices/containers
...
The other thing we talked about here is that we need to confine this test to only when we have a conditional resource...  so we are only checking the function/resources that we can pair... if we cannot pair them, then we won't require the rg parameter.

So we'll need to make sure the "pairing" logic is robust so we don't accidentally exclude something that was there but we just didn't find it.

bmoore-msft (changes requested):

unit-tests/Resources-Should-Not-Be-Ambiguous/Fail/ResourceTypeFirstParameterTrailingSlash.json
```diff
@@ -0,0 +1,13 @@
+﻿{
```
bmoore-msft, line 1: rename this file to be

missing-resourcegroup-parameter-in-resourceId

unit-tests/Resources-Should-Not-Be-Ambiguous/Fail/ResourceTypeTrailingSlash.json
```diff
@@ -0,0 +1,13 @@
+﻿{
+    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
+    "contentVersion": "1.0.0.0",
+
+    "variables": {
+        "trailingslash": "[resourceId(resourceGroup(),'Microsoft.Network/networkSecurityGroups/', 'nsg')]"
```
bmoore-msft: change resourcegroup() to resourceGroup().name

unit-tests/Resources-Should-Not-Be-Ambiguous/Resources-Should-Not-Be-Ambiguous.tests.ps1
```diff
@@ -0,0 +1,6 @@
+﻿
```
bmoore-msft, line 1: we need unit tests for:

child resources unnested:  Microsoft.Network/virtualNetworks/subnets (need one for pass and fail)
child resource nested: "subnets" (and then nested in it's parent) (pass and fail)

bmoore-msft (changes requested):
just need the additional unit tests

PR #265, opened 2020-12-17
Check resource section only to avoid “call depth overflow errors”

tiefpunkt:
We've been observing "The script failed due to call depth overflow." errors on the following three checks:
* adminUsername Should Not Be A Literal
* IDs Should Be Derived From ResourceIDs
* VM Size Should Be A Parameter
 
The errors occurred both in the hosted checks for the Azure marketplace as well as when running them locally:
image
image

I was not able to track down the exact source of the error, but noticed that most other checks focus on a specific section of the template instead of the entire template. That seems to make sense for these three checks as well, so I adjusted them accordingly. Now they go through without any issues and in much less time:
image

Sorry for the whitespace changes, they were done by my text editor automatically.


arm-ttk/testcases/deploymentTemplate/IDs-Should-Be-Derived-From-ResourceIDs.test.ps1
```diff
@@ -15,7 +15,7 @@ $TemplateObject
 )
 
 # First, find all objects with an ID property in the MainTemplate.
-$ids = $TemplateObject  | Find-JsonContent -Key *id -Like
+$ids = $TemplateObject.resources | Find-JsonContent -Key *id -Like
```
bmoore-msft, line 18: can you make sure we have a unit test that will grab a resourceId property from a variable?

arm-ttk/testcases/deploymentTemplate/adminUsername-Should-Not-Be-A-Literal.test.ps1
```diff
@@ -11,15 +11,15 @@ $TemplateObject
 )
 
 # Find all references to an adminUserName
-$adminUserNameRefs = $TemplateObject | 
+$adminUserNameRefs = $TemplateObject.resources |
```
bmoore-msft, line 14: this won't work here because I could have a variable with the same name and while that's not a 100% accurate test, it is something we want to test...

bmoore-msft (changes requested):
See comments - I think we can take pieces of the PR to unblock this (it is a PITA) - we'd also like to resolve the root cause.

Do you have a template/sample that we can use to repro the overflow and debug the underlying cause?  (can file an issue and attach if so)

arm-ttk/testcases/deploymentTemplate/IDs-Should-Be-Derived-From-ResourceIDs.test.ps1
tiefpunkt: The `pass3` resource in `arm-ttk/unit-tests/IDs-Should-Be-Derived-From-ResourceIDs/Pass/exceptions-that-should-pass.json` does that, I think.

arm-ttk/testcases/deploymentTemplate/adminUsername-Should-Not-Be-A-Literal.test.ps1
tiefpunkt: Is it better now?

tiefpunkt:
I do have a template where this currently fails, but it's a marketplace solution template, which I'm not sure I'm allowed to attach here, sorry.

bmoore-msft:
@tiefpunkt - changes look good, would still like to get the repro template - if the marketplace offer is live, then the template code is public... but if you want, just tell me which offer is it and I can go pull it.

tiefpunkt:
@bmoore-msft It went live this week: https://azuremarketplace.microsoft.com/en-us/marketplace/apps/lnw-softgmbh.phoenix-s4hana

Issue #266, opened 2020-12-21
IDs-Should-Be-Derived-From-ResourceIDs false positive on "targetWorkerSizeId"


konjp:
Probably "workerSizeId"(Microsoft.Web/serverFarms) in exceptions needs to change to "targetWorkerSizeId".

bmoore-msft:
@konjp - can you zoom out and give us a JSON snippet for what is failing (i.e. throwing the false positive) - or sounds like if you understand it will enough you could just submit the PR...

NickBishop2112:
Suddenly failing on my local PC for the `"ApplicationId": "[variables('applicationInsightsName')]",`  as per the attached arm code:

```
 {
            "type": "microsoft.insights/components",
            "apiVersion": "2020-02-02-preview",
            "name": "[variables('applicationInsightsName')]",
            "location": "[parameters('location')]",
            "tags": {
                "[concat('hidden-link:', resourceId('Microsoft.Web/sites', variables('applicationInsightsName')))]": "Resource"
            },
            "properties": {
                "ApplicationId": "[variables('applicationInsightsName')]",
                "Request_Source": "IbizaWebAppExtensionCreate"
            }
        }
```
What would cause a change to the regular expression?

The Devops pipeline using the RunARMTTKTests@1 task works fine.

bmoore-msft:
fixed in #267

PR #267, opened 2020-12-21
Update IDs-Should-Be-Derived-From-ResourceIDs.test.ps1

konjp:
https://github.com/Azure/arm-ttk/issues/266


bmoore-msft:
Thanks @konjp - it should be both since the older apiVersions used a different property name, so I just made the tweak.

Issue #182, opened 2020-08-30
Publish to Powershell Gallery


fowl2:
It'd be great if this could be uploaded to PowerShell Gallery so we could `Install-Module` it just like every other cmdlet instead of forking or having to figure out how to download it from github in our devops scripts.

Thanks!

bmoore-msft:
this is planned for this semester

Splaxi:
@bmoore-msft 

Do you have any update on this? Do you need a hand to set it up?

asears:
Any updates on this?  Seems to have been deprioritized?

JustinGrote:
It's not a big deal to publish to the gallery, heck I'll publish it as a pre-release module for you and you can take over the package when you are ready.

bmoore-msft:
@MarcusFelling to see if he can pick this up again...

asears:
How does the installed module work with custom tests, or how can custom tests be injected after import?

bmoore-msft:
Good question... 

@StartAutomating ?

StartAutomating:
@asears
@bmoore-msft 

Looking thru the code, it should be possible to add your own test cases/groups, though not as simple and elegant as we'd like it to be.

Test-AzTemplate -TestCase/-TestGroup will allow you to pass a dictionary of additional test cases and groups.

-TestCase should contain the name/ScriptBlock or Path to test
-TestGroup should contain the list of tests in each group.
Where things get annoying is that while -TestCase is additive, -TestGroup overwrites.  Ergo providing your own test cases _and_ using the built in ones would be annoying.

This could be made more elegant by having something like a -TestDirectory parameter, which would read an additional directory tree's worth of tests (containing it's own /deploymentTemplate and /createUIDefinition subfolders).

I've created Issue #329  to track this specifically.

FallenHoot:
Any update on this? Would be great to have a simple `Install-Module`.

kevinCefalu:
> 
> 
> Any update on this? Would be great to have a simple `Install-Module`.

Seconded! Would be interested in assisting if you are looking for help!

drewkg:
Any update on this? I really need to be able to simply specify the version as just getting the latest is causing my pipelines to fail and I need to manage to update of the module just like any dependency.

Splaxi:
I use this, to download the latest release directly from github, for all my pipelines - while waiting for the publish to powershellgallery.com to happen 🤞

StartAutomating:
We will release on GitHub.

PR #268, opened 2020-12-22
Publish to PowerShell Gallery

MarcusFelling:
To get started on #182 I would like to get arm-ttk published to the PowerShell gallery for more streamlined consumption. 

We will need to identify an [MSA account](https://osgwiki.com//wiki/Internal_Guidelines_for_Adding_to_the_Powershell_Gallery) then generate an API key and add it as a secret in this repo as "NuGetAPIKey".


MarcusFelling:
@bmoore-msft - Do you have access to an MSA account for this?

arm-ttk/arm-ttk.psd1
```diff
@@ -1,6 +1,37 @@
...
+# Generated on: 12/22/2020
+#
+
+@{
+
+    # Script module or binary module file associated with this manifest.
+    RootModule = 'arm-ttk.psm1'
+    
+    # Version number of this module.
+    ModuleVersion = '0.3.#{Patch}#'
```
bmoore-msft, line 15: I think this will cause the local import-module... so I think what we'll do is only publish when this file is changed in a PR... so it has to be intentional, and local import will still work fine.  Keeping in mind that all pipelines today are using this local import approach.  (so this would instantly break those)

.github/workflows/test-publish.yml
```diff
@@ -35,6 +37,11 @@ jobs:
           Import-Module Pester -Force -PassThru -MaximumVersion $PesterMaxVersion} @Parameters
       - name: Check out repository
         uses: actions/checkout@v2
+      - name: Append GitHub run number to patch version in manifest  
+        shell: pwsh
+        run: |   
+          (Get-Content "$env:GITHUB_WORKSPACE\$env:ModulePath") -replace '#{Patch}#', "$env:GITHUB_RUN_NUMBER" `
```
bmoore-msft, line 43: Let's make this an explicit choice or intentional step - rather than pushing every PR to the gallery... this will keep the noise/versions down the the ecosystem isn't as fragmented and also reduce the risk that we instantly break everyone until we mature a bit ;)

.github/workflows/test-publish.yml
```diff
@@ -95,6 +102,13 @@ jobs:
...
-env: 
-  ModulePath: .\arm-ttk\arm-ttk.psd1
+        if: ${{always()}}    
+      - name: Publish module to PowerShell Gallery for master branch pushes only
+        shell: pwsh
+        env:
+          ModuleFolder: arm-ttk
+        run: |
+          Write-Host -Object 'Publishing arm-ttk module to PowerShell Gallery'
+          Publish-Module -Path "$env:GITHUB_WORKSPACE/$env:ModuleFolder" -NuGetAPIKey $env:NuGetAPIKey
```
bmoore-msft, line 112: Here we need an explicit check against the gallery version - if they are the same, don't publish...

arm-ttk/arm-ttk.psd1
```diff
@@ -1,6 +1,37 @@
...
+    
+    # ID used to uniquely identify this module
+    GUID = '20ac342e-df9d-4632-9063-2ab96a9642a8'
+    
+    # Author of this module
+    Author = 'azure-sdk'
```
bmoore-msft, line 21: let's use 'arm-ttk' for the author, the sdk folks are different ;)

arm-ttk/arm-ttk.psd1
```diff
@@ -1,6 +1,37 @@
...
+    
+    # Company or vendor of this module
+    CompanyName = 'Microsoft'
+    
+    # Copyright of this module
+    Copyright = 'Microsoft Corporation. All rights reserved.'
```
bmoore-msft, line 27: need to add a year, e.g. 2021

bmoore-msft (changes requested):

bmoore-msft:
Thanks - I need to go hunt down the key and an MSA...

arm-ttk/arm-ttk.psd1
MarcusFelling: I wasn't sure if we would have to share with sdk or not. arm-ttk it is! :-)

.github/workflows/test-publish.yml
MarcusFelling: I assume we will still want to limit this to master branch only?

.github/workflows/test-publish.yml
MarcusFelling: As mentioned above, versioning will be handled through manifest file updates for now. I have removed this step from the workflow.

arm-ttk/arm-ttk.psd1
MarcusFelling: Ok, i will remove the token place holder. I had the token replacement setup early in the pipeline process to get them to work.  However, you're correct this could hinder local dev, unless a separate dev/local manifest is used. I wasn't sure if you wanted the publish process to be more "automated" or not.  I think manually updating the manifest's version will give us a simple approach to start.

.github/workflows/test-publish.yml
```diff
@@ -1,13 +1,15 @@
-
-name: RunPester
+name: test-publish
```
bmoore-msft, line 1: Let's separate this from the CI/CD pipeline and set this up with a workflow dispatch trigger - that will ensure that publishing is a conscious step and give us time to dogfood a pr if we feel the need. (and not have to reject a PR if we're cautious).

bmoore-msft (changes requested):

MarcusFelling:
Given other priorities, I'm going to close out this PR

PR #278, opened 2021-01-14
95 resource location changes

StartAutomating:


bmoore-msft:
separating into 2 PRs for each scenario

Issue #46, opened 2020-04-27
allow old apiVersion if every newer version is preview


Mikkelcv:
Hi,

I have a case where I use Application Insights (Microsoft.Insights/components), and the possible versions are:
1) 2020-02-02-preview
2) 2018-05-01-preview
3) 2015-05-01
See: https://docs.microsoft.com/en-us/azure/templates/microsoft.insights/allversions

However I am not interested in using preview versions, I prefere the non-preview.
When I run the test of my arm template then it tells me that I am using too old version of this resource:

```
[-] apiVersions Should Be Recent (47 ms)
        Api versions must be the latest or under 2 years old (730 days) - API version 2015-05-01 of Microsoft.Insights/components is 1823 days old
        Valid Api Versions:
        2018-05-01-preview
```
Example:
```json
{
      "name": "[variables('applicationInsightsName')]",
      "type": "Microsoft.Insights/components",
      "apiVersion": "2015-05-01",
      "kind": "web",
      "location": "[parameters('location')]",
      "properties": {
        "Application_Type": "[variables('applicationInsightsApplicationType')]",
        "Flow_Type": "[variables('applicationInsightsFlowType')]",
        "Request_Source": "[variables('applicationInsightsRequestSource')]"
      }
    }
```

Is it possible to configure the tool to ignore preview when there is a non-preview version? 

Is it possible to disallow preview versions completly?

Thanks
Mikkel

bmoore-msft:
thanks @Mikkelcv I thought we had accounted for this but don't see where... good catch

bmoore-msft:
@StartAutomating - I think this is going to be around line 138...

If we are a non-preview version, and all of the later versions are preview - we should not throw

PR #302, opened 2021-02-12
Attempting fix for #46 - Preview api versions should be allowed...

StartAutomating:
if no more recent non-preview apiversions exist.  Attempted fix  of #46


MoneyMan573:
+1 Approve

MoneyMan573:
+1 Approve +1 Approve

Issue #294, opened 2021-02-10
ApiVersion for Microsoft.ApiManagement/service/apis properties


SuperKoelle:
Generally validation of ApiVersion is working but the property object under Microsoft.ApiManagement/service/apis has another value with the same name:

> 

apiVersion | string | No | Indicates the Version identifier of the API if the API is versioned
-- | -- | -- | --

From: [Microsoft Docs](https://docs.microsoft.com/en-us/azure/templates/microsoft.apimanagement/2020-06-01-preview/service/apis#ApiCreateOrUpdateProperties)
This value is also covered by the test but the value here has fewer restrictions and therefore it fails the test with a false negative.

Here is my code:

` {
      "type": "Microsoft.ApiManagement/service/apis",
      "apiVersion": "2020-06-01-preview",
      "name": "[concat(variables('resourceName'), '/', variables('apiName'))]",
      "dependsOn": [
        "[resourceId('Microsoft.ApiManagement/service/apiVersionSets', variables('resourceName'), uniqueString(variables('apiName')))]"
      ],
      "properties": {
        "displayName": "[variables('apiUrlSuffix')]",
        "apiRevision": "1",
        "description": "Init",
        "subscriptionRequired": true,
        "path": "[variables('apiUrlSuffix')]",
        "protocols": [
          "https"
        ],
        "isCurrent": true,
        "apiVersion": "v1",
        "apiVersionSetId": "[resourceId('Microsoft.ApiManagement/service/apiVersionSets', variables('resourceName'), uniqueString(variables('apiName')))]"
      }
    }`

And here is the test result:

> [-] apiVersions Should Be Recent (31 ms)
        Api versions must be a fixed date. Microsoft.ApiManagement/service/apis/ is not.

If I omit:

`"apiVersion": "v1",`

The test passes.

JohnMarkHowell:
Did you get a resolution to this issue? I am seeing it, as well.

SuperKoelle:
Sadly no

bmoore-msft:
Try now - thanks for the repro snippet, never would have guess that one...

SuperKoelle:
Thank you.

It now accepts the apiVersion v1 :)

bmoore-msft:
Cool - thanks for verifying

PR #304, opened 2021-02-19
fix apiVersion and preview version being the same

bmoore-msft:
When an apiVersion has a preview the matches the latest, we don't enforce the latest.  This should fix that...

Also added unit tests, we had very few so hard to track regressions.


bmoore-msft:
#294 - also fixed

PR #306, opened 2021-02-19
Fix broken output formatting in Azure Pipelines.

suic86:
I've changed `arm-ttk/ttk.ezformat.ps1` and regenerated `arm-ttk/arm-ttk.format.ps1xml`.


bmoore-msft:
Excellent - thank you.

Issue #49, opened 2020-04-29
need to account for running tests in nested deployments with scope "inner"


bmoore-msft:
if the expression eval scope is "inner" for a nested deployment, some of the tests won't work properly ( like variable/param not being used) - need to take a inventory of tests to see which need special handling and which do not.

bmoore-msft:
If a template contains a nested deployment where evaluationoptions have scope==inner, then that nested template needs to be tested as it's own template, not as part of the json file it's currently in.  So some workflow could be:

1) parse template
2) does it contain any Microsoft.Resources/deployment where:
```
                "expressionEvaluationOptions": {
                    "scope": "inner"
                },
```
If so, extract the template property and call Test-AzTemplate on those contents (which would recurse becuse this can be nested)

See: https://github.com/bmoore-msft/AzureRM-Samples/blob/master/dynamicSecretId/azuredeploy.inline.json#L25-L45

PR #312, opened 2021-03-05
49-Inner scope

StartAutomating:
Fixing #49 :

1. Adding a regex to find InnerTemplates
2. Updated Parameter/Variable reference tests to account for InnerTemplates
3. Added unit tests

### DIFF
--- a/arm-ttk/Regex/ARM/InnerTemplate.regex.txt
+++ b/arm-ttk/Regex/ARM/InnerTemplate.regex.txt
@@ -1,0 +1,14 @@
+﻿# Matches an inner template
+"template"
+\s{0,}
+:
+\s{0,}
+(?<BalancedCurlyBracket>
+\{                  # An open {
+(?>                 # Followed by...
+    [^\{\}]+|       # any number of non-bracket character OR
+    \{(?<Depth>)|   # an open curly bracket (in which case increment depth) OR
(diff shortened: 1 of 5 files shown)


arm-ttk/Regex/ARM/InnerTemplate.regex.txt
```diff
@@ -0,0 +1,14 @@
+﻿# Matches an inner template
+"template"
```
bmoore-msft, line 2: this needs to account for the expressionEvaluationOptions property (see the issue) - only when it's set to "inner" do we want to change the behavior of the test.

arm-ttk/testcases/deploymentTemplate/Parameters-Must-Be-Referenced.test.ps1
```diff
@@ -18,6 +18,8 @@ param(
 
 $lineBreaks = [Regex]::Matches($TemplateText, "`n|$([Environment]::NewLine)")
 
+$innerTemplates = $TemplateText | ?<ARM_InnerTemplate>
```
bmoore-msft, line 21: add comments

bmoore-msft (changes requested):
A couple notes - also think that we want to think more about an approach as described in the issue (extract and recurse) - that way it works for all tests and we don't have to implement in each one.

Issue #126, opened 2020-07-08
api versions test on locks


tw3lveparsecs:
When using the latest api version for locks on any resource, the api version test fails.

Example of lock on vnet error below.

Api versions must be the latest or under 2 years old (730 days) - API version 2016-09-01 of Microsoft.Network/virtualNetworks/Microsoft.Network/virtualNetworks/providers/locks is 1406 days old in template file

Latest version of locks: https://docs.microsoft.com/en-us/azure/templates/microsoft.authorization/2016-09-01/locks

bmoore-msft:
Can you copy/paste the resource body? I think this is for extension resources, want to make sure (code disambiguates everything)

tw3lveparsecs:
`{
    "type": "Microsoft.Network/virtualNetworks/providers/locks",                                   
    "name": "[concat(parameters('vnetname'), '/Microsoft.Authorization/', parameters('vnetname'), '-lock')]",
    "dependsOn": [
        "[resourceId('Microsoft.Network/virtualNetworks', parameters('vnetname'))]"
    ],
    "apiVersion": "2016-09-01",
    "properties": {
        "level": "CannotDelete"
    }
}`

tw3lveparsecs:
just found another example, if it helps.

Recovery Services Vaults. 
Only has one api available at the moment and api versions test results in a fail. Error below.

Microsoft.PowerShell.Commands.WriteErrorException: Api versions must be the latest or under 2 years old (730 days) - API version 2016-06-01 of Microsoft.RecoveryServices/vaults is 1499 days old in template file.

https://docs.microsoft.com/en-us/azure/templates/microsoft.recoveryservices/2016-06-01/vaults

`{
  "name": "string",
  "type": "Microsoft.RecoveryServices/vaults",
  "apiVersion": "2016-06-01",
  "location": "string",
  "tags": {},
  "identity": {
    "type": "string"
  },
  "properties": {
    "upgradeDetails": {}
  },
  "sku": {
    "name": "string"
  }
}`

bmoore-msft:
Microsoft.RecoveryServices/vaults is actually a correct message - the docs are out of date.  You can see what Azure support by using:

az provider list -o json

The error message from the TTK should give you the supported apiVersions from it's cache

bmoore-msft:
The scope property also needs to be checked...  and example:

https://github.com/Azure/azure-quickstart-templates/blob/master/101-data-share-share-storage-account/azuredeploy.json#L130

[...]

PR #313, opened 2021-03-05
Fixing #126 - Special casing provider resources in apiVersion test

StartAutomating:
Fix for #126 

(special casing provider resources in apiVersion test)


arm-ttk/testcases/deploymentTemplate/apiVersions-Should-Be-Recent.test.ps1
```diff
@@ -86,7 +89,19 @@ foreach ($av in $allApiVersions) {
         $av.type # and adding this resource's type. 
     )
 
-    # To get the full type name, join them all with a slash 
+
+    if ($FullResourceTypes -like '*/providers/*')  # If we have a provider resources
```
bmoore-msft, line 93: if the type has providers, we need to go to the name property to get the names and par that with the type (on line 99) - see the unit test for the proper syntax

bmoore-msft (changes requested):

StartAutomating:
We also want a failure case for provider resources.

PR #315, opened 2021-03-08
fix one issue in resourceId tests (there are more)

bmoore-msft:


asifma:
Thank you for fixing this.
Had issues with e.g: [concat(reference(resourceId('Microsoft.Storage/storageAccounts/', variables('storageAccountName'))).primaryEndpoints.blob, 'vhds/osdisk.vhd')]

bmoore-msft:
yeah - sorry...  I ended up disabling the entire test after a few more issues were found.  So right now this one will give you a false sense of passing.

asifma:
Yeah, E.g. [concat(reference(resourceId('Microsoft.Storage/storageAccounts/', variables('storageAccountName'))).primaryEndpoints.blob, 'vhds/osdisk.vhd')]
is still being extracted to: "Microsoft.Storage/storageAccounts/" which would become an issue.
The last slash needs to be removed as well e.g.

bmoore-msft:
Do you mind adding to #316 as you find things?  Or I guess if you had a list, since the test has been disabled at this point...

PR #358, opened 2021-04-10
bypass resourceId test for bicep generated templates

bmoore-msft:
Adding the ability to skip tests if the template was generated by bicep

the FW won't accept a null or bool for a global var, so have a string for now


arm-ttk/Expand-AzTemplate.ps1
```diff
@@ -144,7 +144,7 @@
                 'CreateUIDefinitionFullPath','createUIDefinitionText','CreateUIDefinitionObject',
                 'FolderName', 'HasCreateUIDefinition', 'IsMainTemplate','FolderFiles',
                 'MainTemplatePath', 'MainTemplateObject', 'MainTemplateText',
-                'MainTemplateResources','MainTemplateVariables','MainTemplateParameters', 'MainTemplateOutputs'
+                'MainTemplateResources','MainTemplateVariables','MainTemplateParameters', 'MainTemplateOutputs', 'TemplateGenerator'
```
bmoore-msft, line 147: change this to pass the entire metadata object - we may also use this for test exculsion (of one form or another).

Issue #361, opened 2021-04-30
Test "IDs-Should-Be-Derived-From-ResourceIDs" is failed for custom query property of ‘Extension/Microsoft_OperationsManagementSuite_Workspace/PartType/LogsDashboardPart’


LeonB87:
I’m creating an Azure Dashboard through ARM. On one of the dashboard items, I’ve created a Query that creates an property called “cycleId”.

This causes ARM-TTK to fail on the “IDs-Should-Be-Derived-From-ResourceIDs”

Snippet from the ARM template: ( I removed some of the code that is long and not necessary)

```json
{
    "position": {
        "x": 8,
        "y": 4,
        "colSpan": 7,
        "rowSpan": 3
    },
    "metadata": {
        "inputs": [
          "removed inputs for brevity"
        ],
        "type": "Extension/Microsoft_OperationsManagementSuite_Workspace/PartType/LogsDashboardPart",
        "settings": {
            "content": {
                "GridColumnsWidth": {
                    "timestamp": "145px",
                    "message": "194px",
                    "customDimensions": "351px",
                    "Type": "172px",
                    "Details": "522px"
                    ** Failing on this CycleId **,
                    "cycleId": "76px",
                    "totalStates": "96px"
                },
                "Query": " *removed beginning part of Query for brevity* | Project timestamp, message, cycleId, totalStates, processed\n",
                "PartTitle": "Policy Processing runs",
                "PartSubTitle": "Cycles"
            }
        },
        "filters": {
            "MsPortalFx_TimeRange": {
                "model": {
                    "format": "local",
                    "granularity": "auto",
                    "relative": "24h"
                }
            }
        }
    }
}
```

The error when running ARM-TTK on the azuredeploy.json file

```text
Errors    : {Property: "cycleId" must use one of the following expressions for an resourceId property:
                        extensionResourceId,resourceId,subscriptionResourceId,tenantResourceId,if,parameters,reference,variables,subscription,guid}
Warnings  : {}
Output    : {}
AllOutput : Property: "cycleId" must use one of the following expressions for an resourceId property:
                        extensionResourceId,resourceId,subscriptionResourceId,tenantResourceId,if,parameters,reference,variables,subscription,guid
Passed    : False
Group     : deploymentTemplate
Name      : IDs Should Be Derived From ResourceIDs
Timespan  : 00:00:00.1678200
File      : {Name, Extension, Text, FullPath…}
```

LeonB87:
Suggested Fix

[...]

PR #362, opened 2021-04-30
Added exclusion for Microsoft.Portal/dashboards

LeonB87:
Added exclusion for Microsoft.Portal/dashboards ID check

Solution for issue #361


LeonB87:
Added unit test with example dashboard that caused the alert to trigger

LeonB87:
And running arm-ttk.tests.ps1 locally completes successfully

Issue #271, opened 2021-01-04
DeploymentTemplate-schema-is-correct test does not check schema property of nested templates


bmoore-msft:
e.g.

```
{
    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
    "contentVersion": "1.0.0.0",
    "resources": [
      {
        "apiVersion": "2020-06-01",
        "name": "nested",
        "type": "Microsoft.Resources/deployments",
        "properties": {
          "mode": "Incremental",
          "template": {
            "$schema": "https://schema.management.azure.com/schemas/2010-01-01/deploymentTemplate.json#",
            "contentVersion": "1.0.0.0",
            "resources": [
            ]
          }
        }
      }
    ]
  }
```

bmoore-msft:
Another case where a recursive approach would catch the instance "for free" - but is independent of expressionEvaluationOptions

bmoore-msft:
@StartAutomating  We may have this, but we need unit tests for inner/outer nested deployments

StartAutomating:
👍

Issue #156, opened 2020-07-09
uris must be created with the uri() function (not concat)


bmoore-msft:
if the property is named "*uri" or "*url" then the value must not start with "[concat("

Skip parameters, but include vars

Also, there are properties that are arrays that we need to check - so if the property type is an array, check each element.

StartAutomating:
@bmoore-msft 👍

PR #363, opened 2021-05-14
Support for Inner templates

### LINKED ISSUE #49: need to account for running tests in nested deployments with scope "inner"

### LINKED ISSUE #139: VM-Images-Should-Use-Latest-Version is still commented out

StartAutomating:
Should fix #49 
May automatically fix #139 #156 #271


bmoore-msft (changes requested):
1) Add some specific tests (in the other unit tests) for nested, inner and outer.  Can essentially duplicate some of the existing test cases as a nested template test case

2) when reporting the error, ideal is to have the line number in the original source that exhibits the error (e.g. line 12+1 (which we're report and the sum (13))

3) we may want to highlight that we are switching to an inner template context but only if this adds something actionable for the user

bmoore-msft:
@VeraBE - can you take a look at this one and make sure it's not going to impact what you consume - there are some output changes.

@StartAutomating can help with any questions...

VeraBE:
@bmoore-msft Thank you for the heads up :) I checked and everything seems to be working properly with this branch

bmoore-msft:
Cool - thx for the look...

Issue #365, opened 2021-05-17
"Location should not be hardcoded" is affected by whether there's a "parameters" object


StephenWeatherford:
The following fails with the "Location Should Not Be Hardcoded" test:

azureDeploy.json:
```json
{
    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
    "contentVersion": "1.0.0.0",
    "parameters": {},
    "resources": [
        {
            "name": "redisCache1",
            "type": "Microsoft.Cache/Redis",
            "apiVersion": "2019-07-01",
            "location": "[resourceGroup().location]",
            "tags": {
                "displayName": "redisCache1"
            },
            "properties": {
                "sku": {
                    "name": "Basic",
                    "family": "C",
                    "capacity": 0
                }
            }
        }
    ]
}
```

But if you remove the "parameters" object, then it passes:

azureDeploy.json:
```json
{
    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
    "contentVersion": "1.0.0.0",
    "resources": [
        {
            "name": "redisCache1",
            "type": "Microsoft.Cache/Redis",
            "apiVersion": "2019-07-01",
            "location": "[resourceGroup().location]",
            "tags": {
                "displayName": "redisCache1"
            },
            "properties": {
                "sku": {
                    "name": "Basic",
                    "family": "C",
                    "capacity": 0
                }
            }
        }
    ]
}
```

Seems to me they should both fail.

bmoore-msft:
fixed in #366 - good catch

StephenWeatherford:
Thanks!  Was really confusing me as to what the rule was supposed to do.

PR #366, opened 2021-05-17
fix 365 and add unit tests

bmoore-msft:
test was not correctly catching the case of using resourceGroup().location and no params were present


bmoore-msft:
#365

PR #375, opened 2021-05-26
added test

bmoore-msft:


jkleppinger:
@bmoore-msft , I'm getting some new exceptions from this added test:
    [-] Secure Params In Nested Deployments (207 ms)
Exception calling "Match" with "1" argument(s): "Value cannot be null.                                               Parameter name: input"  
Seems my $nestedTemplateText is empty for these parameters.  That should be allowed, right?  Should a condition be added to the test so that those resources are skipped or ignored?  Or should I change something in my templates to make this pass?
THANKS!

bmoore-msft:
@jkleppinger - can you open an issue and attach a repro?

PR #382, opened 2021-06-02
Added documentdb / sqldatabases exception

jorgecotillo:
This PR adds an exception to the following type: Microsoft.DocumentDB/databaseAccounts/sqlDatabases but only to resource.id property.

Added a negative test to test that any other id outside of 'resource' property bag should get flagged.


bmoore-msft:
LGTM

PR #401, opened 2021-06-13
#200 - JSONC Support

StartAutomating:
Adding support for JSONC files (Fixing #200 )


unit-tests/_arm-ttk-module/Sample.jsonc
```diff
@@ -0,0 +1,4 @@
+{
```
bmoore-msft, line 1: add a readme.md to this folder to explain that all tests in this folder should "pass"

bmoore-msft (changes requested):

unit-tests/_arm-ttk-module/Sample.jsonc
bmoore-msft: added in #404

PR #412, opened 2021-06-15
334 - Variable Expansion

### LINKED ISSUE #334: add global var for expanded variables to use in running tests

StartAutomating:
Expand-AzTemplate:  Adding .ExpandedTemplateText and .ExpandedTemlateObject Fixes #334


arm-ttk/Expand-AzTemplate.ps1
```diff
@@ -302,7 +303,43 @@ function Expand-AzTemplate
...
+                }
+            }
+
+            if ($expandedTemplateText -ne $TemplateText) {
+                $expandedTemplateObject = try { $expandedTemplateText | ConvertFrom-Json -ErrorAction Stop -ErrorVariable err } catch {
+                    "$_" | Write-Verbose
+                }
+            } else {
+                $expandedTemplateObject = $null
+            }
```
bmoore-msft, line 342: let's add some test cases... a few that come to mind.
- take a resource's properties body and make that properties body an object var - put some violation in that object var


```
"variables": {
  "rg": "[resourceGroup()]"
},
"someProperty": "[variables('rg').location]"

Should also look for the issues that spawned this work...

**Review by bmoore-msft — CHANGES_REQUESTED** [review]

**Review by bmoore-msft — CHANGES_REQUESTED** [review]
When replacing the variable...

If the place we're inserting the variable value startswith [ and endswith ] then remove the brackets from the variable before placing it within

e.g. in this example:
```
{
  "variables": {
    "rg": "[resourceGroup()]"
  },
  "someProperty": "[variables('rg').location]"
}
```

The proper resolution is:
```
"someProperty": "[resourceGroup().location]"
```
not
```
"someProperty": "[[resourceGroup()].location]" //this is bad
```

PR #413, opened 2021-06-15
309 - Test-AzTemplate:  Adding -SkipByFile

StartAutomating:


arm-ttk/Test-AzTemplate.ps1
```diff
@@ -104,6 +104,12 @@ Each test script has access to a set of well-known variables:
     [string[]]
     $Skip,
 
+    # If provided, will skip tests on a file-by-file basis.
```
bmoore-msft, line 107: let's add some examples on how to use and update the docs as well

bmoore-msft (changes requested):

PR #414, opened 2021-06-15
Fixing 104 AllFiles tests

StartAutomating:


arm-ttk/Test-AzTemplate.ps1
```diff
@@ -126,7 +126,7 @@ Each test script has access to a set of well-known variables:
         $builtInTestCases = @{}
         # Next we'll define some human-friendly built-in groups.
         $builtInGroups = @{
-            'all' = 'deploymentTemplate', 'createUIDefinition', 'deploymentParameters'
+            'all' = 'deploymentTemplate', 'createUIDefinition', 'deploymentParameters', 'AllFiles'
```
bmoore-msft, line 129: thinking is that we don't want this by default until we have a better understanding of what may be included in a folder...  we could make it non-default behavior for now to be safe...

bmoore-msft (changes requested):
see comments

bmoore-msft (changes requested):
Need to add test cases

Issue #292, opened 2021-02-03
Hardcoded URI's


drewkg:
When deploying policies, deploy if not exists type, the following is embedded in the ARM

"deployment": {
                "properties": {
                  "mode": "incremental",
                  "template": {
                    "$schema": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
                    "contentVersion": "1.0.0.0",
                    "parameters": {

This is basically a nested ARM template.  This over ever results in a report of

    [-] DeploymentTemplate Must Not Contain Hardcoded Uri 250ms
      RuntimeException: Found hardcoded reference to management.azure.com

Is there a sensible way around the issue or should the test be skipped?

drewkg:
I have spotted the mistake, the original policies that were developed used http://schema.* instead of https://schema.*.

I don't know if this makes a difference, and following the http URI receives a redirect to https with the correct content.

So the question I guess is, was the ARM TTK correct, or should there be a new test / error message for http URI's for allowed hardcoded URI test?

bmoore-msft:
this is similar to #49 where we want to change behavior based on where the "thing" is in the template...  I don't think there's a simple workaround for this specific case, but you can skip the entire test if you don't care about the others (docs have steps for that).

ReneRebsdorf:
@bmoore-msft 
Would you mind explaining why hardcoded URIs are frowned upon?

bmoore-msft:
Primarily because they are fragile... very fragile if the template is used in other clouds (they flat out won't work) and potentially fragile if endpoints need to change.  So for code cleanliness we want to show patterns where you don't have to hardcode them.

StartAutomating:
@bmoore-msft  This should be resolved by the innerTemplates fix.

bmoore-msft:
I think it can/should be - but I don't think we're parsing the template from a policyDefinition resource only from a nested deployment...

So the resource itself will be: ```Microsoft.Authorization/policyDefinitions``` and we need to look for the template property in that resource and treat it as we do a nested deployment.  For a certain type of policy (not all) there will be a deployment property and that will have a template property.

PR #419, opened 2021-06-18
292 Policy definition id inner templates

StartAutomating:
Attempting fix of #292   (this is actually about PolicyDefinition templates).


arm-ttk/Expand-AzTemplate.ps1
```diff
@@ -291,14 +291,14 @@ function Expand-AzTemplate
             
             $innerTemplates = @(if ($templateText -and $TemplateText.Contains('"template"')) {
                 Find-JsonContent -InputObject $templateObject -Key template |
-                    Where-Object { $_.expressionEvaluationOptions.scope -eq 'inner' }
+                    Where-Object { $_.expressionEvaluationOptions.scope -eq 'inner' -or $_.parameters.policyDefinitionID }
```
bmoore-msft, line 294: See: https://docs.microsoft.com/en-us/azure/governance/policy/samples/pattern-deploy-resources#sample-policy-definition for a sample - looks like we want to use ```policyRule```, not ```policyDefinitionId```

bmoore-msft (changes requested):
Also need a test case for it somewhere - be sure to comment in that test case why it's there

arm-ttk/Expand-AzTemplate.ps1
bmoore-msft: Let's add a test case for this specific scenario - the above link should work but need to be wrapped into a policyDefinition resource in a template, e.g.

https://github.com/Azure/azure-quickstart-templates/blob/master/managementgroup-deployments/mg-policy/azuredeploy.json#L37

Issue #385, opened 2021-06-03
false positive in apiVersion test


bmoore-msft:
Template is below, not sure if it's a cache problem or a nested template problem... (due to changes for inner scope eval)

```
{
    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
    "contentVersion": "1.0.0.210602",
    "parameters": {
        "location": {
            "type": "string",
            "defaultValue": "[resourceGroup().location]",
            "metadata": {
                "description": "The location in which the resources should be deployed."
            }
        },
        "Source": {
            "type": "string",
            "metadata": {
                "description": "URL of a ZIP archive containing VHD images files. For example: https://provider.com/bin/support/azure-vhd.zip"
            }
        },
        "Destination": {
            "type": "string",
            "metadata": {
                "description": "URI of a Storage Account container with a writable SAS Token. For example: https://blob-storage-name.blob.core.windows.net/container-name?sp=racwl&st=2021-01-01T12:00:00&se=2021-02-01T12:00:00&spr=https&sv=2021-01-01&sig=123412341234"
            }
        }
    },
    "variables": {
        "urlVHDZipArchive": "[parameters('Source')]",
        "uriWritableStorageAccountBlobContainerSasToken": "[parameters('Destination')]",
        "nestedTemplateName_ImportVHD-FromZipArchive": "[concat('ImportVHD-FromZipArchive-',uniqueString(parameters('Source'),uniqueString(parameters('Destination'))))]"
    },
    "resources": [
        {
            "type": "Microsoft.Resources/deployments",
            "apiVersion": "2021-04-01",
            "name": "[variables('nestedTemplateName_ImportVHD-FromZipArchive')]",
            "properties": {
                "mode": "Incremental",
                "expressionEvaluationOptions": {
                    "scope": "inner"
                },
                "parameters": {
                    "location": {
                        "value": "[parameters('location')]"
                    },
                    "Source": {
                        "value": "[variables('urlVHDZipArchive')]"
                    },
                    "Destination": {
                        "value": "[variables('uriWritableStorageAccountBlobContainerSasToken')]"
                    },
                    "deploymentScriptName": {
                        "value": "[variables('nestedTemplateName_ImportVHD-FromZipArchive')]"
                    }
                },
                "template": {
                    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
                    "contentVersion": "1.0.0.210602",
[...]

PR #421, opened 2021-06-18
355 symbolic naming tests

StartAutomating:


unit-tests/Secure-Params-In-Nested-Deployments/Pass/Nested-Outer-Nothing-SymbolicNaming.json
```diff
@@ -0,0 +1,55 @@
+﻿{
```
bmoore-msft, line 1: let's add a failure case with this same json - it will test a nested template via symbolic name - the failure case will ensure we can find something that deep.

bmoore-msft (changes requested):

unit-tests/Secure-Params-In-Nested-Deployments/Pass/Nested-Outer-Nothing-SymbolicNaming.json
bmoore-msft: Grab the template from #385 it has a nested deployment... add another resource to it at the top level and convert to symbolic names.  Then probably a pass and failure case using each.

Issue #316, opened 2021-03-08
resources should not be ambiguous test has multiple issues


bmoore-msft:
This file seems to be a good test case on many fronts, sooo....

```json
{
    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
    "contentVersion": "1.0.0.0",
    "parameters": {
      "Location": {
        "type": "string",
        "defaultValue": "[resourceGroup().location]",
        "metadata": {
          "description": "Location to deploy resources to."
        }
      },
      "AppName": {
        "type": "string",
        "defaultValue": "[concat('FuncApp-', uniqueString(resourceGroup().id))]",
        "metadata": {
          "description": "Name of the Function App."
        }
      },
      "plan": {
        "type": "string",
        "defaultValue": "[concat(parameters('AppName'), '-asp')]",
        "metadata": {
          "description": "Name of the Elastic Premium Plan for the Function App."
        }
      },
      "skuName": {
        "type": "string",
        "defaultValue": "EP1",
        "metadata": {
          "description": "The Sku name/size of the Elastic Premium plan"
        }
      },
      "vnetName": {
        "type": "string",
        "defaultValue": "VirtualNetwork",
        "metadata": {
          "description": "Name of the VNET that the Function App and Storage account will communicate over."
        }
      },
      "functionsSubnetName": {
        "type": "string",
        "defaultValue": "FunctionsSubnet",
        "metadata": {
          "description": "The subnet that the Function App will use for VNET traffic."
        }
      },
      "storagePESubnetName": {
        "type": "string",
        "defaultValue": "StoragePESubnet",
        "metadata": {
          "description": "The subnet that will be used for the Storage private endpoints."
        }
      },
      "StorageAccountName": {
        "type": "string",
        "defaultValue": "[toLower(concat(uniqueString(resourceGroup().id), 'pe'))]",
        "metadata": {
          "description": "Name of the Storage account that the Function App will use for operations and content."
        }
      },
      "virtualNetworkAddressPrefix": {
        "type": "string",
        "defaultValue": "10.100.0.0/16",
        "metadata": {
          "description": "VNET address space."
        }
      },
      "functionSubnetAddressPrefix": {
        "type": "string",
        "defaultValue": "10.100.0.0/24",
        "metadata": {
          "description": "Function App's subnet address range."
        }
      },
      "privateEndpointSubnetAddressPrefix": {
        "type": "string",
[...]

PR #426, opened 2021-06-21
Attempting fix for #316:  

StartAutomating:
Changing check for any foundResource with condition (mutliple finds had caused this to always return true)


bmoore-msft:
There is an issue in Find-JsonContent that needs to be cleaned up before we can merge this...

bmoore-msft:
Let's add a test case (or two or three) - also we should re-enable the test from #316 to ensure this fix also fixed that test.

Issue #437, opened 2021-06-23
Blocks bicep modules (inner scoped nested templates): False positives for "Template Should Not Contain Blanks"


StephenWeatherford:
```sh
   [-] Template Should Not Contain Blanks (7 ms)                                                                                                                        Empty property:  {} found on line: 25 Line: 25, Column: 20     
```
See failure in pipeline: https://dev.azure.com/azurequickstarts/azure-quickstart-templates/_build/results?buildId=72429&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a&t=14b22f05-782e-55fa-3e59-b66c82f514a9&l=116

shorter repro:
```json
{
  "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
  "contentVersion": "1.0.0.0",
  "resources": [
    {
      "type": "Microsoft.Resources/deployments",
      "apiVersion": "2019-10-01",
      "name": "bastion",
      "properties": {
        "expressionEvaluationOptions": {
          "scope": "inner"
        },
        "mode": "Incremental",
        "template": {
          "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
          "contentVersion": "1.0.0.0",
          "metadata": {
            "_generator": {
              "name": "bicep",
              "version": "0.4.1.14562",
              "templateHash": "13967474456372168665"
            }
          },
          "parameters": {
            "location": {
              "type": "string",
              "defaultValue": "[resourceGroup().location]",
              "metadata": {
                "description": "Specifies the location of AKS cluster."
              }
            },
            "bastionHostName": {
              "type": "string",
              "metadata": {
                "description": "Specifies the name of the Azure Bastion resource."
              }
            },
            "bastionSubnetId": {
              "type": "string",
              "metadata": {
                "description": "Specifies the id of the Azure Bastion subnet resource id."
              }
            }
          },
          "functions": [],
          "variables": {
            "bastionPublicIpAddressName": "[format('{0}PublicIp', parameters('bastionHostName'))]"
          },
          "resources": [
            {
              "type": "Microsoft.Network/publicIPAddresses",
              "apiVersion": "2020-08-01",
              "name": "[variables('bastionPublicIpAddressName')]",
              "location": "[parameters('location')]",
              "sku": {
                "name": "Standard"
              },
              "properties": {
                "publicIPAllocationMethod": "Static"
              }
            },
            {
[...]

Issue #443, opened 2021-07-01
JSon Parsing Error, nested template


drewkg:
When I embed an empty nested template I receive teh following error

ConvertFrom-Json: C:\Users\ARM\Source Code\GitHub\drewkg\Azure\ARM\Policy\DiagnosticSettings\LogAnalytics\ARM-TTK\arm-ttk\Expand-AzTemplate.ps1:304:51                                                                                                                                                 LLine |                                                                                                                                                304 |                  $TemplateObject = $TemplateText | ConvertFrom-Json                                                                                |                                                    ~~~~~~~~~~~~~~~~                                                                                | Conversion from JSON failed with error: Additional text encountered after finished reading JSON content: :. Path '', line 1,                       | position 10. 

The nested section, taken from a Bicep repository, converted into ARM, is as follows

  "resources": [
    {
      "name": "getManagementGroupName",
      "type": "Microsoft.Resources/deployments",
      "apiVersion": "2019-10-01",
      "location": "[parameters('location')]",
      "properties": {
        "expressionEvaluationOptions": {
          "scope": "inner"
        },
        "mode": "Incremental",
        "template": {
          "$schema": "https://schema.management.azure.com/schemas/2019-08-01/managementGroupDeploymentTemplate.json#",
          "contentVersion": "1.0.0.0",
          "resources": []
        }
      }
    },

This provides the ability to retrieve the current Management Group name that the deployment is being run against.

nsucheninov:
The same issue. We have some ARM templates that contains nested templates. ARM TTK generates two errors for these templates:
`Conversion from JSON failed with error: Additional text encountered after finished reading JSON content: :. Path '', line 1, position 10."`
`Empty property:  {} found on line: 50"`
A sample template is attached.
[applicationInsights.zip](https://github.com/Azure/arm-ttk/files/6753947/applicationInsights.zip)

drewkg:
I am not sure if this is related, but this template https://raw.githubusercontent.com/drewkg/Azure/feature/ARMTTK/ARM/Policy/DiagnosticSettings/LogAnalytics/azuredeploy.json

is still causing issues when parsing the json, the error message from AzDO pipeline is ##[error]ConvertFrom-Json : Invalid JSON primitive:  {}{

bmoore-msft:
@drewkg - can you open a new issue for this one?  Feel free to cross ref this one...

drewkg:
[...]

PR #455, opened 2021-07-16
Fixing #443 : Removing 'template': from start of $TemplateText

StartAutomating:


StartAutomating:
Root Cause Analysis:  This was creating an invalid JSON fragment, which impacted Location-Should-Not-Be-Hardcoded.  This test converted $TemplateText back from JSON to create an in-memory copy.

StartAutomating:
This also may fix #437

Issue #439, opened 2021-06-23
Blocks bicep modules (inner scoped nested templates): Method exception: Cannot find an overload for "Matches" and the argument count: "2". 


StephenWeatherford:
```sh
   [+] IDs Should Be Derived From ResourceIDs (10 ms)                                                                                                           MethodException: /Users/stephenweatherford/repos/arm-ttk/arm-ttk/Test-AzTemplate.ps1:319                                                                         Line |                                                                                                                                                            319 |                              $InnerTemplateStartLine =                                                                                                         |                              ~~~~~~~~~~~~~~~~~~~~~~~~~                                                                                                         | Cannot find an overload for "Matches" and the argument count: "2".                                                                                                                                                                                             
```
Failure in pipeline: https://dev.azure.com/azurequickstarts/azure-quickstart-templates/_build/results?buildId=72429&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a&t=14b22f05-782e-55fa-3e59-b66c82f514a9&l=116

simplier repro:
```json
{
  "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
  "contentVersion": "1.0.0.0",
  "resources": [
    {
      "type": "Microsoft.Resources/deployments",
      "apiVersion": "2019-10-01",
      "name": "bastion",
      "properties": {
        "expressionEvaluationOptions": {
          "scope": "inner"
        },
        "mode": "Incremental",
        "template": {
          "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
          "contentVersion": "1.0.0.0",
          "parameters": {
            "location": {
              "type": "string",
              "defaultValue": "[resourceGroup().location]",
              "metadata": {
                "description": "Specifies the location of AKS cluster."
              }
            }
          },
          "resources": [
            {
              "type": "Microsoft.Network/publicIPAddresses",
              "apiVersion": "2020-08-01",
              "name": "[variables('bastionPublicIpAddressName')]",
              "location": "[parameters('location')]",
              "sku": {
                "name": "Standard"
              },
              "properties": {
                "publicIPAllocationMethod": "Static"
              }
            }
          ]
        }
      }
    }
  ]
}
```

StephenWeatherford:
Goes away if you change expressionEvaluationOptions scope to 'outer'

bmoore-msft:
#385 also has a repro

[...]

PR #456, opened 2021-07-16
Test-AzTemplate:  Fixing #439.  

StartAutomating:
Updating Logic to Detect InnerTemplate ranges (was unable to find start when more than one failure presented from a test)


bmoore-msft:
Let's grab the test cases from #439 and #385, whittle them down to min repro and add for this...

Issue #345, opened 2021-04-07
 Allowed Values Should Actually Be Allowed show errors also for location/allowedValues


danielecazzari:
In my createUiDefinition.json I'm defining the allowed location as shown below.
This seems to be correct according to documentation but arm-ttk report this error

 [-] Allowed Values Should Actually Be Allowed (61 ms)                                                                       
CreateUIDefinition parameter  with value eastus is not allowed in the main template parameter createOrUpdate            CreateUIDefinition parameter  with value eastus2 is not allowed in the main template parameter createOrUpdate           CreateUIDefinition parameter  with value southcentralus is not allowed in the main template parameter  
.....

I'm wondering if config/basics tags are confused with "basics" steps by scripts.

Regards,

Daniele

My code:
            "basics": { 
                "location": {
                    "allowedValues": [
                        "eastus",
                        "eastus2",
                        "southcentralus",
                        "westus2",
                        "australiaeast",
                        "southeastasia",
                        "northeurope",
                        "uksouth",
                        "westeurope",
                        "centralus",
                        "northcentralus",
                        "westus",
                        "southafricanorth",
                        "centralindia",
                        "eastasia",
                        "japaneast",
                        "koreacentral",
                        "canadacentral",
                        "francecentral",
                        "germanywestcentral",
                        "uaenorth",
                        "brazilsouth",
                        "centralusstage",
                        "eastusstage",
                        "eastus2stage",
                        "northcentralusstage",
                        "southcentralusstage",
                        "westusstage",
                        "westus2stage",
                        "asia",
                        "asiapacific",
                        "australia",
                        "brazil",
                        "canada",
                        "europe",
                        "global",
                        "india",
                        "japan",
                        "uk",
                        "unitedstates",
                        "eastasiastage",
                        "southeastasiastage",
                        "centraluseuap",
                        "eastus2euap",
                        "australiasoutheast",
                        "japanwest",
                        "southindia",
                        "westindia",
                        "canadaeast",
                        "ukwest",
                        "uaecentral",
                        "brazilsoutheast"                        
                    ]
                }
            }

bmoore-msft:
@danielecazzari - can you attach an entire file with a repro? (original or a paired down version)

danielecazzari:
Hi,

I'm sorry for the long waiting, but I had to find time to develop a paired down version that you could find here:
[createUiDefinition.json.txt](https://github.com/Azure/arm-ttk/files/6304934/createUiDefinition.json.txt)

[...]

PR #484, opened 2021-07-30
424 allowed values fix

### LINKED ISSUE #424: Unclear Allowed Values Should Actually Be Allowed error
Hi,

I'm receiving the below error but I'm not really able to understand what I did wrong.
Could you help me?

I could share my template via mail if needed.  

 CreateUIDefinition
  [-] Allowed Values Should Actually Be Allowed (2848 ms)
  CreateUIDefinition parameter  is not allowed in the main template parameter clusterName
  CreateUIDefinition parameter postgreSQLTier is not allowed in the main template parameter postgreSQLserverName
  CreateUIDefinition parameter managedClustersDropDown is not allowed in the main template parameter clusterName
  CreateUIDefinition parameter userAssignedIdentitiesDropDown is not allowed in the main template parameter clusterName  

Regards,

Daniele

StartAutomating:
Should fix #424 and #345

### DIFF
--- a/arm-ttk/testcases/CreateUIDefinition/Allowed-Values-Should-Actually-Be-Allowed.test.ps1
+++ b/arm-ttk/testcases/CreateUIDefinition/Allowed-Values-Should-Actually-Be-Allowed.test.ps1
@@ -35,9 +35,9 @@ foreach ($av in $allowedValues) { # Walk thru each thing we find.
     $p = $count * 100/ $allowedValues.Count
     Write-Progress -Id $progressId "Checking Controls $($controlName)" " " -PercentComplete $p
     $stepName = $av.ParentObject[1].name # If the grandparent object has a name field, we're in steps
-    $lookingFor= if ($stepName) { "*steps(*$stepName*).$controlName*"} else {"*basics(*$($controlName)*"} 
+    $lookingFor= if ($stepName) { $stepName } else { $controlName } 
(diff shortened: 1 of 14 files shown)


arm-ttk/testcases/CreateUIDefinition/Allowed-Values-Should-Actually-Be-Allowed.test.ps1
```diff
@@ -35,9 +35,9 @@ foreach ($av in $allowedValues) { # Walk thru each thing we find.
     $p = $count * 100/ $allowedValues.Count
     Write-Progress -Id $progressId "Checking Controls $($controlName)" " " -PercentComplete $p
     $stepName = $av.ParentObject[1].name # If the grandparent object has a name field, we're in steps
-    $lookingFor= if ($stepName) { "*steps(*$stepName*).$controlName*"} else {"*basics(*$($controlName)*"} 
+    $lookingFor= if ($stepName) { $stepName } else { $controlName }
```
bmoore-msft, line 38: I think $stepName needs to look similar to before the change, i.e. step+control - since we're looking for a control not a step. e.g. steps('stepName').controlName

We should also make sure we have a test case (or testcases) that cover

steps('stepName').controlName
steps('stepName').sectionName.controlName

bmoore-msft (changes requested):

Issue #487, opened 2021-08-04
JSON Parse gives a strange warning and causes GitHub Super Linter to fail and can't be skipped


spoelly:
Back from vacation we see that the ARM-TTK linter is not working fine in our Super Linter execution.
After running the ARM-TTK locally we see this strange warning for one of our templates:

VERBOSE: Conversion from JSON failed with error: After parsing a value an unexpected character was encountered: o. Path 'variables.kvtAccessPoliciesArray', line 118, position 97.

After some debugging we see that line 118 is not the issue but this one:

''' 
"mergeKvtAccessPoliciesCondition": "[not(contains(string(parameters('kvtAccessPolicies')),concat('\"objectId\":\"',parameters('kvtObjectId'),'\"')))]",
'''

Very strange because the json is valid otherwise the deployment of our template would fail. So as a work-arround we tried to skip the "JSONFiles Should Be Valid" check but this is imposible. :( 

**This is our JSON file**

```json

{
  "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
  "contentVersion": "1.0.0.0",
  "parameters": {
    "kvtName": {
      "type": "string",
      "metadata": {
        "description": "Name of the new Keyvault"
      }
    },
    "kvtSkuName": {
      "type": "string",
      "defaultValue": "Standard",
      "allowedValues": [
        "Standard",
        "Premium"
      ],
      "metadata": {
        "description": "SKU for the vault"
      }
    },
    "kvtEnabledForDeployment": {
      "type": "bool",
      "defaultValue": false,
      "metadata": {
        "description": "Set TRUE to enable access for deployments"
      }
    },
    "kvtEnabledForTemplateDeployment": {
      "type": "bool",
      "defaultValue": false,
      "metadata": {
        "description": "Set TRUE to enable access for Template deployments"
      }
    },
    "kvtEnabledForDiskEncryption": {
      "type": "bool",
      "defaultValue": false,
      "metadata": {
        "description": "Set TRUE to enable access for disk encryption"
      }
    },
    "kvtObjectId": {
      "type": "string",
      "defaultValue": "",
      "metadata": {
        "description": "ObjectId of Pipeline account"
      }
    },    
    "kvtNetworkAclsBypass": {
      "type": "string",
      "defaultValue": "None",
      "allowedValues": [
        "None",
        "AzureServices"
      ],
      "metadata": {
        "description": "Tells what traffic can bypass the network rules"
      }
    },
[...]

Issue #483, opened 2021-07-30
failure to parse json on bicep generated file


bmoore-msft:
see: https://dev.azure.com/azurequickstarts/azure-quickstart-templates/_build/results?buildId=78536&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a&t=14b22f05-782e-55fa-3e59-b66c82f514a9&l=13

comes from this file: 

https://github.com/Azure/azure-quickstart-templates/blob/master/managementgroup-deployments/create-subscription-resourcegroup/azuredeploy.json

StephenWeatherford:
Here's another example: https://github.com/Azure/azure-quickstart-templates/pull/11403

Related to nested templates?

```json
  {
    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
    "contentVersion": "1.0.0.0",
    "metadata": {
      "_generator": {
        "name": "bicep",
        "version": "0.4.412.5873",
        "templateHash": "8360920427510730960"
      }
    },
    "parameters": {
      "_artifactsLocation": {
        "type": "string",
        "defaultValue": "[deployment().properties.templateLink.uri]",
        "metadata": {
          "description": "The base URI where artifacts required by this template are located including a trailing '/'"
        }
      },
      "_artifactsLocationSasToken": {
        "type": "secureString",
        "defaultValue": "",
        "metadata": {
          "description": "The sasToken required to access _artifactsLocation.  When the template is deployed using the accompanying scripts, a sasToken will be automatically generated. Use the defaultValue if the staging location is not secured."
        }
      },
      "location": {
        "type": "string",
        "defaultValue": "[resourceGroup().location]",
        "metadata": {
          "description": "Location for all resources."
        }
      },
      "HostPublicIPAddressName": {
        "type": "string",
        "defaultValue": "HVHOSTPIP",
        "metadata": {
          "description": "Resource Name for Public IP address attached to Hyper-V Host"
        }
      },
      "virtualNetworkName": {
        "type": "string",
        "defaultValue": "VirtualNetwork",
        "metadata": {
          "description": "Hyper-V Host and Guest VMs Virtual Network"
        }
      },
      "virtualNetworkAddressPrefix": {
        "type": "string",
        "defaultValue": "10.0.0.0/22",
        "metadata": {
          "description": "Virtual Network Address Space"
        }
      },
      "NATSubnetName": {
        "type": "string",
        "defaultValue": "NAT",
[...]

Issue #468, opened 2021-07-23
JSon Parsing, Nested Template


drewkg:
When I embed the following resource

    {
      "name": "getManagementGroupName",
      "type": "Microsoft.Resources/deployments",
      "apiVersion": "2019-10-01",
      "location": "[parameters('location')]",
      "properties": {
        "expressionEvaluationOptions": {
          "scope": "inner"
        },
        "mode": "Incremental",
        "template": {
          "$schema": "https://schema.management.azure.com/schemas/2019-08-01/managementGroupDeploymentTemplate.json#",
          "contentVersion": "1.0.0.0",
          "resources": []
        }
      }
    },

Into a management group template, when running ARM TTK locally I get

  [-] Error occurred in test script '.\ARM\Policy\DiagnosticSettings\LogAnalytics\arm-ttk-tests.ps1' 250ms                                                                                  JsonReaderException: Additional text encountered after finished reading JSON content: :. Path '', line 1, position 10.                                                                  ArgumentException: Conversion from JSON failed with error: Additional text encountered after finished reading JSON content: :. Path '', line 1, position 10.                            at Expand-AzTemplate<Process>, C:\Users\drewk\Source Code\GitHub\drewkg\Azure\ARM-TTK\arm-ttk\Expand-AzTemplate.ps1: line 305                                                           at Test-AzTemplate<End>, C:\Users\drewk\Source Code\GitHub\drewkg\Azure\ARM-TTK\arm-ttk\Test-AzTemplate.ps1: line 626                                                                   at <ScriptBlock>, C:\Users\drewk\Source Code\GitHub\drewkg\Azure\ARM\Policy\DiagnosticSettings\LogAnalytics\arm-ttk-tests.ps1: line 1                                                   at <ScriptBlock>, C:\Users\drewk\OneDrive\Documents\PowerShell\Modules\Pester\4.1.0\Pester.psm1: line 825                                                                               at Invoke-Pester<End>, C:\Users\drewk\OneDrive\Documents\PowerShell\Modules\Pester\4.1.0\Pester.psm1: line 840                                                                          at <ScriptBlock>, C:\Users\drewk\Source Code\GitHub\drewkg\Azure\Utilities\Run-ARMTTK.ps1: line 9                                                                                       at <ScriptBlock>, C:\Users\drewk\Source Code\GitHub\drewkg\Azure\Utilities\Run-ARMTTK.ps1: line 8                                                                                       at <ScriptBlock>, <No file>: line 1  

[...]

PR #488, opened 2021-08-04
Resolve-JSONContent:  Fixes for Inner Template

StartAutomating:
To avoid comparison issues.

This was causing the TargetIndex to be incorectly compared, which in turn caused Resolve-JSONContent to fail to find the match, which in turn caused the template to fail to snip at the right locations (and thus become corrupted)


StartAutomating:
This also appears to fix #487

StartAutomating:
This also fixes #483

bmoore-msft:
and #468

PR #512, opened 2021-09-10
Additional unit tests

StartAutomating:


unit-tests/Allowed-Values-Should-Actually-Be-Allowed/Pass/LocationValueAllowed/mainTemplate.json
```diff
@@ -0,0 +1,22 @@
+{
+    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "location": {
+            "type": "string",
+            "allowedValues": [
+                "test"
```
bmoore-msft, line 8: this test  is in the pass folder but as written should fail...

And the pipeline didn't seem to care.

bmoore-msft (changes requested):

PR #516, opened 2021-10-02
Fix for when latest stable API version is greater than two years old

jakelevinez:
First PR on open source, happy to get feedback and make changes.

### DIFF
--- a/arm-ttk/testcases/deploymentTemplate/apiVersions-Should-Be-Recent.test.ps1
+++ b/arm-ttk/testcases/deploymentTemplate/apiVersions-Should-Be-Recent.test.ps1
@@ -153,6 +153,7 @@ foreach ($av in $allApiVersions) {
     # Create a string of recent or allowed apiVersions for display in the error message
     $recentApiVersions = ""
 
+    #add latest stable apiVersion to acceptable list by default
     $stableApiVersions = $validApiVersions | where-object { $_ -notmatch 'preview' } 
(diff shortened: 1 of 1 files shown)


arm-ttk/testcases/deploymentTemplate/apiVersions-Should-Be-Recent.test.ps1
```diff
@@ -208,14 +219,19 @@ foreach ($av in $allApiVersions) {
             $trimmedApiVersion = $validApiVersions[0].ToString().Substring(0, $validApiVersions[0].ToString().LastIndexOf("-"))
             $nonPreviewVersionInUse = ($trimmedApiVersion -eq $av.apiVersion)
         }
-        if (-not $nonPreviewVersionInUse) {
+        if (-not $nonPreviewVersionInUse) {            
+            if ($($av.ApiVersion) -eq $latestStableApiVersion) {
```
bmoore-msft, line 223: Let's remove this Write-Warning - I'm thinking of the scenario where the latest version is more than 2 years old, i.e. there's nothing newer - which is not terribly uncommon.  In that case I wouldn't want to warn.

There is a case where you're using the latestStable and there are new preview versions but I think even in those cases that's intentional so would not want to warn there.  Basicially, we don't need to encourage the use of a preview version even if the one being used is really old.

jakelevinez:
@bmoore-msft Updated for the reference function test as well, and added a testcase for it.

Warning has been removed.

PR #522, opened 2021-10-06
Set rule IDs

VeraBE:
Define rule IDs on best practices for deployment

It will be consumed by BPA: https://github.com/Azure/template-analyzer/pull/183


StartAutomating:
@VeraBE 

The TargetObject is used for reporting the line/column/piece of JSON that failed.

Thus we cannot change the TargetObject without changing things that use it for reporting.

Is there any reason we cannot use the .ErrorID ?  We already provide an -ErrorID for all errors that I'm aware of.  It should already be unique.  Please help us understand the goal here.

VeraBE:
@bmoore-msft 
> * some of the tests have the same ID is that intentional?

Do you mean that different test files have the same ID or that each file has different types of errors and I'm using the same ID? The first one would be unintentional but I can't find them, and about the second: One of the reasons to have these types of IDs is that we want users to be able to map to the corresponding documentation easily from a message in the console https://github.com/Azure/template-analyzer/blob/verabe/ruleIDs/docs/built-in-bpa-rules.md At first I did a different ID for each different type of error even in the same file, but then I thought it would be confusing when looking at the documentation, considering that the TTK docs are currently split basically by the file: https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/template-test-cases (Oh, just noticed our docs don't have the proper titles matching with the TTK docs, I can fix that)

> * when I add a test, how do I know what ID/number to use?

By looking at https://github.com/Azure/template-analyzer/blob/verabe/ruleIDs/docs/built-in-bpa-rules.md Definety not ideal :/ In the last meeting, I understood that the plan was to not add new checks in TTK, so I thought it shouldn't be a problem for now, and when it gets fully integrated into BPA it also should be less annoying

> * I added a test earlier this week I assume it needs an idea as well? https://github.com/Azure/arm-ttk/blob/master/arm-ttk/testcases/deploymentTemplate/Password-params-must-be-secure.test.ps1

Thanks, I'll fix that

@StartAutomating 
> The TargetObject is used for reporting the line/column/piece of JSON that failed.
> Thus we cannot change the TargetObject without changing things that use it for reporting.

Is there another tool consuming TTK that breaks when new properties are added to that object?
I didn't like to use this object at first but this usage doesn't seem to be that far off from the intention of the object: https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/write-error?view=powershell-7.1 "Specifies the object that was being processed when the error occurred", we could think that those specific rules were being processed when the error was found

> Is there any reason we cannot use the .ErrorID ? We already provide an -ErrorID for all errors that I'm aware of. It should already be unique. Please help us understand the goal here.

We want to have a consistent ID between these checks and the ones in TemplateAnalyzer (BPA), and that's the format we decided on for BPA. To use the existing ErrorIds we would need to keep a map somewhere between these IDs and BPA IDs. Seems to be more cumbersome considering that when editing/adding a rule here we would need to make a change there. Besides this solution is only temporal until TTK gets fully integrated into BPA. Editing the ErrorId properties to have our format also doesn't seem to work out because those are being used in tests here and I think that changing the logic of those tests wouldn't be especially simple

VeraBE:
We decided to not support rule IDs for PowerShell checks until those are run natively in BPA

PR #535, opened 2021-11-05
On branch master: add inner template test case for Outputs-Must-Be-Present-In-Template-Parameters

galiacheng:
Error happens for `Outputs-Must-Be-Present-In-Template-Parameters` test case, when the main template has inner template, and the inner template does not have parameters the same from the UiDefinition outputs.  It's super common usage. Such case should pass.

Error message shows like:
```
 [-] Outputs Must Be Present In Template Parameters (6209 ms)
        output acrName does not exist in template.parameters
        output aksAgentPoolNodeCount does not exist in template.parameters
        output aksAgentPoolVMSize does not exist in template.parameters
        output aksClusterName does not exist in template.parameters
        output aksClusterRGName does not exist in template.parameters
```

### DIFF
--- a/arm-ttk/Test-AzTemplate.ps1
+++ b/arm-ttk/Test-AzTemplate.ps1
@@ -1,4 +1,4 @@
-﻿function Test-AzTemplate
+function Test-AzTemplate
 {
     [Alias('Test-AzureRMTemplate')] # Added for backward compat with MP
     <#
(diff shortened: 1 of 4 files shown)


bmoore-msft (commented):
Thanks for the fix - we think a more general fix is in order for this one, e.g. the test should never be called -- and there are some other tests that are being called even though they shouldn't be.  It may take a few days to fix that so I'm going to merge this to unblock and then we may roll back some of these once that's done.

galiacheng:
Thank you @bmoore-msft

PR #550, opened 2021-12-03
325 blank property exception improvement

StartAutomating:


bmoore-msft (changes requested):
We want to allow blanks in any descendent, not just direct.  So in the appSettings property on a web app, anything of any shape in that property can be empty.
https://github.com/Azure/azure-quickstart-templates/blob/master/quickstarts/microsoft.web/webapp-windows-aspnet/azuredeploy.json#L45

That works when it's a property.

The other "change" would be to exempt a property (And it's descendents) when it's of a certain type, e.g.

"type": "Microsoft.Web/sites/config"

we can exempt everything under the "properties" property for that given type.  The generic feature is just combining the descendants capability along with a type+property name exemption.

bmoore-msft:
This needs to take into account the resourceType... so we can say something like:

property "bar" can be blank when found under a resourceType "Microsoft.Resources/blah"

PR #575, opened 2022-01-25
Adding a marketplace mode for ttk and including new tests for external URIs

raosuhas:
This PR does 2 things : 

- Adds a marketplace mode for ttk. This allows us to run the tests in a mode that reduces some of the warnings to errors so that some of the cases that are not relevant for marketplace can be made optional.


bmoore-msft:
@raosuhas - can you separate these into 2 PRs?  One for the new tests and one for the new mode?

bmoore-msft:
@StartAutomating - can you take a look at the one for the new mode?  I think we might have a simpler way to do it?

raosuhas:
Ok .. I am keeping this as the PR for adding the mode. Will submit a separate one for the test. 

---
In reply to: [1022698082](https://github.com/Azure/arm-ttk/pull/575#issuecomment-1022698082) [](http://example.com/codeflow?ancestors=1022698082,1022698082)

raosuhas:
Let me know if there is a simpler way. Unfortunately the decision where the tests is to be made into a error or warning is only done by the Write-Error Command which is why i needed to abstract that away

---
In reply to: [1023698536](https://github.com/Azure/arm-ttk/pull/575#issuecomment-1023698536) [](http://example.com/codeflow?ancestors=1023698536,1023698536)

Issue #583, opened 2022-02-04
Exception calling "Matches" with "2" argument(s) when recursing quickstart test folder


bmoore-msft:
Running the TTK over the test folder in the quickstart repo shows the error below.  To repro put an *empty template* azuredeploy.json file in the test folder and run the command:

```Test-AzTemplate -templatePath ./azuredeploy.json/```
or
```Test-AzTemplate test #from the parent folder```

The following appear multiple times - the multiple times part may be a function of #581 

```
MethodInvocationException: C:\Users\bmoore\source\repos\arm-ttk\arm-ttk\Test-AzTemplate.ps1:332
Line |
 332 |                              $InnerTemplateStartLine =
     |                              ~~~~~~~~~~~~~~~~~~~~~~~~~
     | Exception calling "Matches" with "2" argument(s): "Start index cannot be less than 0 or greater than input length. (Parameter 'startat')"

```

azuredeploy.json contents:
```
{
  "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
  "contentVersion": "1.0.0.0",
  "parameters": { },
  "variables": { },
  "resources": [ ],
  "outputs": { }
}
```

bmoore-msft:
s/b fixed in #589

Issue #581, opened 2022-02-03
apiVersions Should Be Recent is duplicating error with nested deployments


bmoore-msft:
repro template below - the list of errors below is repeated multiple times showing different lines for the same errors.

```
    [-] apiVersions Should Be Recent (511 ms)
        The apiVersion 2021-02-01 was not found for the resource type: Microsoft.Network/virtualNetworks/subnets
        Api versions must be the latest or under 2 years old (730 days) - API version 2020-01-01 of Microsoft.Network/privateDnsZones is 764 days old Line: 286, Column: 20127
        Valid Api Versions:
        2020-06-01
        2020-06-01
        Api versions must be the latest or under 2 years old (730 days) - API version 2020-01-01 of Microsoft.Network/privateDnsZones/virtualNetworkLinks is 764 days old
        Valid Api Versions:
        2020-06-01
        2020-06-01
        Api versions must be the latest or under 2 years old (730 days) - API version 2020-01-01 of Microsoft.Network/privateDnsZones is 764 days old Line: 633, Column: 47369
        Valid Api Versions:
        2020-06-01
        2020-06-01
        Api versions must be the latest or under 2 years old (730 days) - API version 2020-01-01 of Microsoft.Network/privateDnsZones/virtualNetworkLinks is 764 days old Line: 864, Column:
65523
        Valid Api Versions:
        2020-06-01
        2020-06-01
        The apiVersion 2020-06-01 was not found for the resource type: Microsoft.Network/privateEndpoints/privateDnsZoneGroups
        Api versions must be the latest or under 2 years old (730 days) - API version 2020-01-01 of Microsoft.Network/privateDnsZones is 764 days old Line: 444, Column: 31945
        Valid Api Versions:
        2020-06-01
        2020-06-01
        Api versions must be the latest or under 2 years old (730 days) - API version 2020-01-01 of Microsoft.Network/privateDnsZones/virtualNetworkLinks is 764 days old Line: 579, Column:
42671
        Valid Api Versions:
        2020-06-01
        2020-06-01
        Api versions must be the latest or under 2 years old (730 days) - API version 2020-01-01 of Microsoft.Network/privateDnsZones is 764 days old
        Valid Api Versions:
        2020-06-01
        2020-06-01
        Api versions must be the latest or under 2 years old (730 days) - API version 2020-01-01 of Microsoft.Network/privateDnsZones/virtualNetworkLinks is 764 days old Line: 953, Column:
73849
        Valid Api Versions:
[...]

Issue #580, opened 2022-02-03
apiVersions Should Be Recent - is showing duplicate apiVersion in the valid versions list


bmoore-msft:
```
    [-] apiVersions Should Be Recent (511 ms)
        The apiVersion 2021-02-01 was not found for the resource type: Microsoft.Network/virtualNetworks/subnets
        Api versions must be the latest or under 2 years old (730 days) - API version 2020-01-01 of Microsoft.Network/privateDnsZones is 764 days old Line: 286, Column: 20127
        Valid Api Versions:
        2020-06-01
        2020-06-01
        Api versions must be the latest or under 2 years old (730 days) - API version 2020-01-01 of Microsoft.Network/privateDnsZones/virtualNetworkLinks is 764 days old
        Valid Api Versions:
        2020-06-01
        2020-06-01
```

https://github.com/Azure/azure-quickstart-templates/blob/master/quickstarts/microsoft.machinelearningservices/machine-learning-end-to-end-secure/azuredeploy.json

bmoore-msft:
Still repros after #589

StartAutomating:
@bmoore-msft  looks good now

Issue #582, opened 2022-02-03
Parameters Must Be Referenced - shows errors for parameters that aren't present and incorrectly flags others in nested templates


bmoore-msft:
we seem to combine params from multiple  templates (or the output) but the params we flag are indeed used.  So some are false positives and some are not even defined in the template they are flagged in.

https://github.com/Azure/azure-quickstart-templates/blob/master/quickstarts/microsoft.machinelearningservices/machine-learning-end-to-end-secure/azuredeploy.json

bmoore-msft:
s/b fixed in #589

PR #589, opened 2022-02-09
Inner templates improvements

StartAutomating:
* Find-JsonContent:  Fixing Underlying issue of #583 (2nd or more JSONPath in the same hierarchy would be returned corrupted)
* Returning one result per inner template node (this cleans up reporting) (#580, #581)
* Accepting return objects with a .JSONPath for easier line association
* Adding exception handling around [Matches] (#583)
* Adding additional context to Test-AzTemplate return value


bmoore-msft:
Let's add tests from #580 #581 #582 since we expect to knock those out with this...

arm-ttk/Resolve-JSONContent.ps1
```diff
@@ -256,7 +256,7 @@
                 Line     = [Regex]::new('(?>\r\n|\n|\A)', 'RightToLeft').Matches(
                                 $JSONText, $indexMatch.Index
                            ).Count
-                Column   = $listMatch.Groups["ListItem"].Index + $(
+                Column   = $listMatch.Groups["ListItem"].Index - $(
```
bmoore-msft, line 259: Let's add a test case for this change...

PR #592, opened 2022-02-15
Add test for artifacts location

raosuhas:
Adds a test for artifacts location


bmoore-msft:
@raosuhas - do we still need this?

PR #597, opened 2022-02-23
Enhancing the password box validation

v-rucdu:


bmoore-msft:
@v-rucdu - can you provide some detail for the change?  i.e. why do we need it?

Issue #319, opened 2021-03-09
Passing Secrets across deployments should use secureString properties


bmoore-msft:
When using nested deployments and working with parameters that contain secrets, we should ensure the parameters being used in the nested template are secureStrings.

bmoore-msft:
This will require fetching the template (whether inline or relativePath) and looking at the param type

StartAutomating:
@bmoore-msft could you provide a succinct example or two?

bmoore-msft:
TLDR; parameter types should match between parent and child template:

```json
{
    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
    "contentVersion": "1.0.0.0",
    "parameters": {
        "secret": {
            "type": "secureString"
        }
    },
    "resources": [
        {
            "name": "nestedDeployment1",
            "type": "Microsoft.Resources/deployments",
            "apiVersion": "2021-04-01",
            "properties": {
                "expressionEvaluationOptions": {
                    "scope": "inner"
                },
                "mode": "Incremental",
                "parameters": {
                    "secret": {
                        "value": "[parameters('secret')]"
                    }
                },
                "template": {
                    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
                    "contentVersion": "1.0.0.0",
                    "parameters": {
                        "secret": {
                            "type": "string"
                        }
                    },
                    "variables": {},
                    "resources": [],
                    "outputs": {}
                }
            }
        }
    ],
    "outputs": {}
}
```

bmoore-msft:
- For a given template look for all deployments resources
- for each deployment resource fetch the parameters that are passed into that template (not the ones defined but the ones passed
- look up the type of each parameter passed from the parent template
- now fetch the type of the parameter in the deployment resource template definition.

The two should match

bmoore-msft:
via #607

PR #607, opened 2022-03-11
Fixing #319 parameter consistency

StartAutomating:
Fixing #319 and adding unit tests


bmoore-msft:
Added detail to #319 for what we think is it the right way to do this...

Issue #618, opened 2022-03-24
Old Location-Should-Not-Be-Hardcoded Test is no longer valid


StartAutomating:
We have a [test](unit-tests\Location-Should-Not-Be-Hardcoded\Fail\inner-scope-resource.json) that no longer causes errors, because we are no longer checking deploymentTemplates due to changes in bicep behavior.

For the moment, I believe we simply want to remove the test.

Unfortunately, the test highlights a scenario we could do a better job of checking:   that an actual hardcoded value exists for a resource group.

If you would like to change the behavior of Location-Should-Not-Be-Hardcoded to accomplish this, please let me know.

PR #619, opened 2022-03-24
Removing inner-scope-resource (Fixing #618)

StartAutomating:


bmoore-msft:
This looks to be a valid test case are we sure it should be removed?

bmoore-msft:
Conversation recap:

We don't seem to be checking the location property for resources - probably missed this since this is one of the oldest tests.

For the location property on each resource, the property value must be one of the following:
- global
- any expression

Note that resourceGroup().location, is an expression but we have tests in place to check for that.

Issue #622, opened 2022-03-31
Always have an error "VM Size Should Be A Parameter"


uzurpator1:
We have faced the issue with this test works with big nested template. It is always warning that VM Size Should Be A Parameter, but it already in parameter.
[primaryBackend.zip](https://github.com/Azure/arm-ttk/files/8390574/primaryBackend.zip)

Steps to reproduce:

Extract files from the attached archive
Run Test-AzTemplate -TemplatePath "your-path-to-the-extracted-files\primaryBackend.json" -Test deploymentTemplate -ErrorAction Continue
Observe the error:
[-] VM Size Should Be A Parameter (57 ms)                                                                           
        VM Size must be a parameter 

vmSize parameter already set in parameters

bmoore-msft:
The error I'm seeing on your attached template is from the ```Deploy_NLB_``` resource... on that resource the parameters section has:

```json
          "vmSize": {
            "type": "string",
            "defaultValue": "Standard_D2s_v3"
          }
```
in the parameters section of the ```/deployments``` resource.  That's not valid - it's a param definition but you need the param value, i.e.
```json
          "vmSize": {
            "value": "..."
          }
```

So the TTK is throwing an error on it, that it's terribly helpful aside from the fact it's catching a deployment failure.  Also, if the other ```/deployments``` resource wasn't there, you probably would have figured out the problem.

I'm going to repurpose the bug though - so we add some information about where the error is found.

bmoore-msft:
added resourceType and name to the error message via #628

PR #628, opened 2022-04-01
improved error message for vm size failure

bmoore-msft:
since we just check for any property named "vmSize" anywhere in the template, adding some info to the error message about which resource contains the error.


bmoore-msft:
For #622

PR #653, opened 2022-06-10
added update check

bmoore-msft:


bmoore-msft:
@StartAutomating

StartAutomating:
@bmoore-msft : I'd do this in the module load.  Putting it in each Test-AzTemplate would -a- be slow and -b- skew metrics.

bmoore-msft:
Ah, yes...  good idea...  how would we do that? ;)

StartAutomating:
@bmoore-msft looks like you've got it in the right place now.

Issue #667, opened 2022-07-27
ARM-TTK should skip Bicep self-generated variable


galiacheng:
When using Bicep [loadTextContent()](https://docs.microsoft.com/en-us/azure/azure-resource-manager/bicep/bicep-functions-files#loadtextcontent)  function, `bicep build` generates a variable (e.g.  `$fxv#0`) to save the file content and references it in the template.

ARM-TTK complains the Bicep self-generated variable saying: `Unreferenced variable: $fxv#0`, which is not expected.

### How to repro

Bicep version: 0.8.9

Sample bicep file `test.bicep`:

```bicep
param identity object = {}
param location string
param utcValue string = utcNow()

resource deploymentScript 'Microsoft.Resources/deploymentScripts@2020-10-01' = {
  name: 'test'
  location: location
  kind: 'AzureCLI'
  identity: identity
  properties: {
    azCliVersion: '2.33.1'
    scriptContent: loadTextContent('./test.sh')
    cleanupPreference: 'OnSuccess'
    retentionInterval: 'P1D'
    forceUpdateTag: utcValue
  }
}
```

Build the script with `bicep build test.bicep`, it will generate a json file `test.json`.
Use ARM-TTK to validate `test.json`: `Test-AzTemplate -TemplatePath test.json`, you will find an error saying `Unreferenced variable: $fxv#0`.

### Workaround

The workaround was figured out by @majguo.

Use `loadFileAsBase64()` to load the file, and call `base64ToString()` to parse base64 string.

```bicep
param identity object = {}
param location string
param utcValue string = utcNow()

var base64_test=loadFileAsBase64('./test.sh')

resource deploymentScript 'Microsoft.Resources/deploymentScripts@2020-10-01' = {
  name: 'test'
  location: location
  kind: 'AzureCLI'
  identity: identity
  properties: {
    azCliVersion: '2.33.1'
    scriptContent: base64ToString(base64_test)
    cleanupPreference: 'OnSuccess'
    retentionInterval: 'P1D'
    forceUpdateTag: utcValue
  }
}
```

bmoore-msft:
@StartAutomating - looks like a bug in the regex in irregular - doesn't seem to handle the `#`

StartAutomating:
@bmoore-msft 

Your suspicion was close to correct.  It didn't like #, because that can start a Regex comment if not escaped.

It also didn't like $, because that normally is interpreted as an "end anchor" if not escaped.

PR should fix both cases.

[...]

PR #669, opened 2022-07-29
Allowing -Variable to contain dollar sign and number sign ( Fixing #667 )

StartAutomating:
Fixing #667

### DIFF
--- a/unit-tests/Variables-Must-Be-Referenced/Pass/dollar.json
+++ b/unit-tests/Variables-Must-Be-Referenced/Pass/dollar.json
@@ -1,0 +1,15 @@
+{
+    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
+    "contentVersion": "1.0.0.0",
+    "variables": {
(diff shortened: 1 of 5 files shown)


unit-tests/Variables-Must-Be-Referenced/Pass/dollar.json
```diff
@@ -0,0 +1,15 @@
...
+    "contentVersion": "1.0.0.0",
+    "variables": {
+      "f$": "#bin/bash\necho 'hi'"
+    },
+    "resources": [
+    ],
+    "outputs": {
+      "test": {
+        "type": "string",
+        "value": "[variables('f$')]"
```
bmoore-msft, line 12: this unit test is failing....

StartAutomating: @bmoore-msft - that makes sense, the Regex had only been allowing $ as the first character.

bmoore-msft: ty

Issue #635, opened 2022-04-12
[parameters('anyParameter')[copyIndex()]] is incorrectly treated as an array and throws type mismatch error


sahitya-bhandarkar:
I am passing an object parameter to the nested template example array[i] (with copyIndex in picture).
The ARM ttk test thinks of this parameter as an array.
 
The bicep files I have mentioned below are for reference only. I am running the ARM ttk **Test-AzTemplate** on the the compiled json file.

---------------------------------------------------------------
moduletest.bicep
---------------------------------------------------------------

module roleAssignments './roleAssignments.bicep' = [for (roleAssignment,index) in roleAssignmentCollection: {
  name: 'roleAssignments-${index}'
  scope: resourceGroup(roleAssignment.resourceGroupName)
  params: {
    roleAssignment: roleAssignment
  }
  
---------------------------------------------------------------
roleAssignments.bicep
---------------------------------------------------------------
param roleAssignment object

var storageAccountRole = subscriptionResourceId('Microsoft.Authorization/roleDefinitions', roleAssignment.roleId)

resource storageRoleAssignment 'Microsoft.Authorization/roleAssignments@2020-10-01-preview' = {
 .
 .
 .
 
 
 ---------------------------------------------------------------
 az bicep build - converted json
 ---------------------------------------------------------------
 .
 .
 .
 {
  "copy": {
	"name": "roleAssignments",
	"count": "[length(parameters('roleAssignmentCollection'))]"
  },
  "type": "Microsoft.Resources/deployments",
  "apiVersion": "2020-10-01",
  "name": "[format('roleAssignments-{0}', copyIndex())]",
  "resourceGroup": "[parameters('roleAssignmentCollection')[copyIndex()].resourceGroupName]",
  "properties": {
	"expressionEvaluationOptions": {
	  "scope": "inner"
	},
	"mode": "Incremental",
	"parameters": {
	  **"roleAssignment": {
		"value": "[parameters('roleAssignmentCollection')[copyIndex()]]"
	  },**
	  "webAppPrincipalId": {
		"value": "[reference(resourceId('Microsoft.Web/sites', parameters('webAppName')), '2021-01-15', 'full').identity.principalId]"
	  },
	  "webAppName": {
		"value": "[parameters('webAppName')]"
	  }
	},
	"template": {
	  "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
	  "contentVersion": "1.0.0.0",
	  "metadata": {
[...]

PR #672, opened 2022-08-10
Fix#635 : Referenced Parameters incorrectly treated as type mismatch

yks0000:
Fix https://github.com/Azure/arm-ttk/issues/635


yks0000:
@bmoore-msft Added test case. Could you please help review it.

PR #691, opened 2022-09-26
Adding false positive case to Template Should Not Contain Blanks 

dciborow:
Adding False Positive Test, to check if this issue is current or related to marketplace version.


dciborow:
#692

Issue #697, opened 2022-10-04
API Version check fails if API is older than 2 years, even if it is the latest version


sam-cogan:
It seems that test to check for API versions, at least for some resources, are failing when the latest version availible is older than 2 years. An example of this is Microsoft.Resources/deploymentScripts where the latest version is 2020-10-01 and there is no newer version.

Based on my reading of the test, if there is no newer version then the test should pass as long as you are using the latest version. I think the error is down to this code in the test:

```   
if (($timeSinceApi.TotalDays -gt $NumberOfDays) -and ($howOutOfDate -gt 0)) {
        # if the used apiVersion is the second in the list, check to see if the first in the list is the same preview version (due to sorting)
        # for example: "2017-12-01-preview" and "2017-12-01" - the preview is sorted first so we think we're out of date
        $nonPreviewVersionInUse = $false
        if ($howOutOfDate -eq 1) {
            $trimmedApiVersion = $validApiVersions[0].ToString().Substring(0, $validApiVersions[0].ToString().LastIndexOf("-"))
            $nonPreviewVersionInUse = ($trimmedApiVersion -eq $av.apiVersion)
        }
        if (-not $nonPreviewVersionInUse) {            
            if ($($av.ApiVersion) -eq $latestStableApiVersion) {     
                #break from loop to avoid throwing error when using latest stable API version           
                break
            }

            # If it's older than two years, and there's nothing more recent
            Write-Error "Api versions must be the latest or under $($NumberOfDays / 365) years old ($NumberOfDays days) - API version $($av.ApiVersion) of $FullResourceType is $([Math]::Floor($timeSinceApi.TotalDays)) days old" -ErrorId ApiVersion.OutOfDate -TargetObject $av
            Write-Output "Valid Api Versions:`n$recentApiVersions"
        }
    }
```

I think the ```-not $nonPreviewVersionInUse``` part is leading to a double negative which means it will only work if the latest version is a preview version, rather than what is intended, for it not to be a preview version. I will submit a PR to fix this, but feel free to correct me if I have the wrong end of the stick

bmoore-msft:
Thanks Sam - I've pushed #699 to ignore bicep module apiVersions

PR #698, opened 2022-10-04
Fix issue with latest api version being older than 2 years

sam-cogan:
Remove ```-not``` which seems to require that the api version be a preview version, if the latest version is older than 2 years old, which I believe to be the opposite of the intention.


sam-cogan:
#697

PR #699, opened 2022-10-04
added exception for old bicep version

bmoore-msft:
bicep modules generate an old apiVersion and the user cannot change it - adding an exception for /deployments resources when bicep is detected.


majguo:
Hello @bmoore-msft I observed a similar issue today and documented the details in #704. Could you pls also help on it? Cc @edburns.

PR #720, opened 2022-12-05
Update Password-params-must-be-secure.test.ps1

dciborow:
Bool Parameter should be excluded from this check.


dciborow:
Well... could not reproduce so must be an issue with my current version.

bmoore-msft (changes requested):
looks good - can you add a unit test?

PR #734, opened 2023-03-22
Updated Regex used by Textbox Length Check Test

### LINKED ISSUE #731: Incorrect Regex (Length Check) Validation
### Description

When using `arm-ttk` to validate `createUiDefination.json` file the regex validation check is flagging regex's that have length's specified as not having length condition.

### Validation Result

Regex Validation

### UI Element Definition

```json
{
	"name": "objectId",
	"type": "Microsoft.Common.TextBox",
	"label": "Object Id",
	"toolTip": "Id of User or Service Principal that can access secrets from Key Vault",
	"constraints": {
		"required": true,
		"validations": [
			{
				"regex": "^[0-9a-fA-F]{8}-(?:[0-9a-fA-F]{4}-){3}[0-9a-fA-F]{12}$",
				"message": "Invalid Object Id Pattern"
			}
		]
	},
	"visible": true
},
{
	"name": "name",
	"type": "Microsoft.Common.TextBox",
	"label": "Data Factory Name",
	"toolTip": "Only supported characters can be used in name",
	"constraints": {
		"required": true,
		"validations": [
			{
				"regex": "^(?!-)[A-Za-z0-9-]{3,63}(?<!-)$",
				"message": "Only alphanumeric characters are allowed, and value must be 3-36 characters long"
			}
		]
	},
	"visible": true
}
```

Regex's being incorrectly flagged

```
^[0-9a-fA-F]{8}-(?:[0-9a-fA-F]{4}-){3}[0-9a-fA-F]{12}$
^(?!-)[A-Za-z0-9-]{3,63}(?<!-)$
```

### Observations

For the 2nd regex on removing the negative lookbehind the validation passes. Possibly the current check only looks at the end of the regex to see if a length has been provided. As for the 1st regex event though it ends with a length it is still flagged as a warning.

dvdmtw98:
fixes #731

### Current Code

- The test only checks to see if the length constraint has been provided at the start and end of the regex string but it is possible for the length constraint to be present anywhere in the string. 
- Additionally the current test only returns success if length constraint of the pattern `{m,n}` or `{m,}` is provided while there is a third valid option `{m}` which will return failure.

### Updated Code

- Since we only need to check if length constraint has been provided somewhere in the input regex string there is no need to check the start and end of the string separately instead a search of the entire string should be sufficient.
- Added support for the `{m}` variant of length constraint which was not currently supported.
- Updated the Units Tests with additional test conditions to check a more broader range of possible inputs.

Additional test cases that where validated can be found here: https://regex101.com/r/fde0fy/3

### DIFF
--- a/arm-ttk/testcases/CreateUIDefinition/Textboxes-Are-Well-Formed.test.ps1
+++ b/arm-ttk/testcases/CreateUIDefinition/Textboxes-Are-Well-Formed.test.ps1
@@ -13,8 +13,8 @@ param(
 
 # First, find all textboxes within CreateUIDefinition.
 $allTextBoxes = $CreateUiDefinitionObject | Find-JsonContent -Key type -value Microsoft.Common.TextBox
-# Regex Constraint should contain the pattern {m,n} or {m,} or {m} (Length Check)
-$lengthContraintRegex = [Regex]::new('{\d+(?:,\d+|,)?}')
+# Regex Constraint should contain one of following pattern (Length Check): {m,n} or {m,} or {m}
+$lengthContraintRegex = [Regex]::new('\{(?<Min>\d+)(?:,(?<Max>\d+)|,)?\}')
 
 foreach ($textbox in $allTextBoxes) {
     # Then we walk over each textbox.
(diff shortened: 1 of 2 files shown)


dvdmtw98:
@microsoft-github-policy-service agree

PR #735, opened 2023-04-05
Create Outputs-Cannot-Contain-ListKeys-Variables.json

dciborow:
Add example of case that is expected to fail.

@bmoore-msft, per the discussion over [here](https://github.com/Azure/bicep/discussions/9293) seems that in Bicep, when the output of listKeys is assigned to a variable, and then an output, no alert is triggered. But, I thought I had seen ARM-TTK properly catch this in the past.... Proposing this test case to confirm or deny my memory.

If this test case passes, we could merge this PR in as confirmation. But if it fails, I expect the fix would be done on the bicep side.


dciborow:
Oops, looks like ARM-TTK won't catch this case after all.

psah434:
Output cannot have secrets.

Issue #733, opened 2023-03-01
EditableGrid Schema requirement does not match ARM-TTK requirement


msalemcode:
The [public publish schema](https://learn.microsoft.com/en-us/azure/azure-resource-manager/managed-applications/microsoft-common-editablegrid?WT.mc_id=Portal-Microsoft_azure_createUIDef) for EditableGrid UI does not required ToolTip field,

However ARM-TTK throw error if toolTip is missing. I notice EditableGrid is not part of e[xception UI here ](https://github.com/Azure/arm-ttk/blob/master/arm-ttk/testcases/CreateUIDefinition/Tooltips-Should-Be-Present.test.ps1)

The question now is, Should EditableGrid be part of the exception list since the published schema shows it is not required? or Should the public schema get fix and in this case could you share where is the schema you using as reference? It should be a public schema.

vidapour:
This change has been merged. Thank you!

PR #739, opened 2023-05-11
EditableGrid does not need tooltips

psah434:
Related to issue https://github.com/Azure/arm-ttk/issues/733
 EditableGrid does not require tooltips in the public schema and this should be followed for toolkit also.
Additional test for the same has been added.


raosuhas (approved):
:shipit:

Issue #737, opened 2023-04-25
Failing to validate Azure Front Door ARM property


danielmacedo90:
**Description**

The `arm-ttk`  is failing to handle **patternsToMatch** property referring the **Azure Front Door**.

**Output / Error**

![image](https://user-images.githubusercontent.com/32306990/234383557-ab1560f9-90e2-4bd9-9a9f-56fa7ac859ca.png)

**ARM Template Definition**

```json
"routingRules": [
          {
            "name": "routingRule1",
            "properties": {
              "frontendEndpoints": [
                {
                  "id": "[resourceId('Microsoft.Network/frontDoors/frontendEndpoints', variables('frontDoorName'), 'frontendEndpoint1')]"
                }
              ],
              "acceptedProtocols": [
                "Https"
              ],
              "patternsToMatch": [
                "/*"
              ],
              "routeConfiguration": {
                "@odata.type": "#Microsoft.Azure.FrontDoor.Models.FrontdoorForwardingConfiguration",
                "forwardingProtocol": "HttpsOnly",
                "backendPool": {
                  "id": "[resourceId('Microsoft.Network/frontDoors/backendPools', variables('frontDoorName'), 'backendPool1')]"
                }
              },


              "enabledState": "Enabled"
            }
          },
]
```

vidapour:
@danielmacedo90 can you provide your full template so that we can test your scenario? Thanks :)

danielmacedo90:
@vidapour here is the full template.

[azuredeploy.zip](https://github.com/Azure/arm-ttk/files/12022524/azuredeploy.zip)

PR #755, opened 2023-10-27
Changing convertfrom-json utility usage based on powershell version (powershell 6+ has utility for json comments)

psah434:
Fixing issue https://github.com/Azure/arm-ttk/issues/737
Running the template provided in the above issue, if using the custom ConvertFrom-JSON utility, it fails.
The code change here changes the utility used to be the default powershell version if the powershell major version is equal to or above 6.
There is a small fix in the convertfrom-json utility where for recognizing comments both the starting and ending characters are searched.

### DIFF
--- a/arm-ttk/arm-ttk.psm1
+++ b/arm-ttk/arm-ttk.psm1
@@ -1,5 +1,6 @@
 ﻿#region JSON Functions
-if ($PSVersionTable.PSEdition -ne 'Core') {
+#powershell 6+ provides support for json with comments
+if ($PSVersionTable.PSVersion.Major -lt 5) {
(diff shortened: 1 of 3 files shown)


arm-ttk/arm-ttk.psm1
```diff
@@ -1,5 +1,6 @@
 ﻿#region JSON Functions
-if ($PSVersionTable.PSEdition -ne 'Core') {
+#powershell 6+ provides support for json with comments
```
raosuhas, line 2: could you please update the description ? I am unableto understand the change made here ?

raosuhas: Lets please update the descriptions for these ...Its really hard to figure out context on these PRs later

raosuhas (approved):

PR #760, opened 2024-01-18
Added property requiredDataConnectors in Template-Should-Not-Contain-Blanks Test to skip if []

v-amolpatil:
Ignore "requiredDataConnectors" property if [].

After running tests locally we see below for the added test case on "requiredDataConnectors" property :
image


v-amolpatil:
> Why do you need the property to be empty? Is there documentation regarding this property being empty public?

Please refer below screenshot and is publicly available:

image

PR #761, opened 2024-02-08
Added property serializedData in Template-Should-Not-Contain-Blanks T…

v-rusraut:
Ignore "serializedData" if it contain [] property.
After running tests locally we see below for the added test case on "serializedData" property :
image

### DIFF
--- a/arm-ttk/testcases/deploymentTemplate/Template-Should-Not-Contain-Blanks.test.ps1
+++ b/arm-ttk/testcases/deploymentTemplate/Template-Should-Not-Contain-Blanks.test.ps1
@@ -41,7 +41,8 @@ param(
     # A value of "properties.settings" would allow blanks in any subproperty of settings.
     [Collections.IDictionary]
     $ResourcePropertiesThatCanBeEmpty = @{
-        "Microsoft.Web/sites/config" = "properties"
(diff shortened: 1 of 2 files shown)


psah434:
Why should the property be empty? Is there public documentation regarding this? Also the test needs to change since serizliedData is not empty in the unit test

v-rusraut:
Here, we are not validating whether the serializedData property is empty. Instead, we are validating within serializedData to check if the property "Headers" are empty ([]). If so, we want to skip this property to avoid ARM TTK failures.
To resolve this ARM TTK error, we must manually remove this property from serializedData. Therefore, we believe it should not result in an ARM TTK failure.

psah434:
Why should the property be empty? Is there public documentation regarding this?

Also if you want serializedData.Headers should be empty, can you change the code accordingly?

v-rusraut:
Here source file contains headers and columns property with []. That's why serializedData contains headers and columns property with []. so, we need to skip this validation.

image

Please refer documentation link: https://learn.microsoft.com/en-us/azure/azure-monitor/visualize/workbooks-automate#work-with-json-formatted-workbook-data-in-the-serializeddata-template-parameter

Thanks

PR #767, opened 2024-03-01
Added Microsoft.Solutions.GraphApiControl to the no-ToolTip controls

raouche:
Updated Tooltips-Should-Be-Present.test.ps1 to add Microsoft.Solutions.GraphApiControl to the list of controls that do not need to be checked for a ToolTip since Microsoft.Solutions.GraphApiControl has no UI.


raouche:
@bmoore-msft Hi Brian - Is this repo (arm-ttk) still being actively maintained? If not, what's the best way to get someone's attention to issues and PR's? Thanks!

PR #778, opened 2024-05-20
Bug: Missing dot-import for Format-Json

luis-gasparschroeder:
In order for Format-Json to be automatically included when importing the arm-ttk module, Format-Json needs to be dot-imported in the arm-ttk.psm1.


StartAutomating:
@luis-gasparschroeder please kindly give this function a more specific name.

"Format-JSON" is too broad of a name to use, for a module that is not generally related to JSON formatting.

Additionally, it is not "formatting", is it "converting".

I would strongly recommend something along the lines of:

`ConvertFrom-TTK` (though `Format-TTK` might also work)

